Issue #5, opened 2014-07-29
GitHub requests return '302 Found', notifications not displayed


dmihal:
I'm using the extension with GitHub Enterprise

The extension displays the error icon (![image](https://cloud.githubusercontent.com/assets/1661138/3740738/83b9c444-1758-11e4-8646-eec010749967.png)), and debugging shows that GitHub is returning 302 Found with an empty response.

![image](https://cloud.githubusercontent.com/assets/1661138/3740766/c01d6b48-1758-11e4-823b-c0effc874a4d.png)

I'm not sure if there's anything that can be done about this or if this is just on GitHub's end...

PR #14, opened 2015-01-08
grant XHR to any URL for GitHub Enterprise

bigwheel:
hello!

I encounterd almost same problem #5 while Im using private ghe.
Of cource, I set correct notification URL.

I took a look a bug and found chrome caused an error about 'Allow-Control-Allow-Origin'.
I thought need more permission for ghe URL.
https://github.com/sindresorhus/github-notifier-chrome/blob/master/extension/manifest.json#L16

I clone this extension and rewrite "https://github.com/" to "_://_/".
Now works right now.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -13,7 +13,7 @@
 	"permissions": [
 		"alarms",
 		"activeTab",
-		"https://github.com/"
+		"*://*/"
 	],
 	"optional_permissions": [
 		"*://*/"


sindresorhus:
That's already in optionalDependencies https://github.com/sindresorhus/github-notifier-chrome/pull/14/files#diff-2ccaee6c5cfa6c1e0a36166894df97bcR19 and will be asked on-demand if you use a custom URL.

bigwheel:
I understand optional_permissions just now.
However, anyway github-notifier-chrome doesnt work in ghe.
From Step3 in https://developer.chrome.com/extensions/permissions, I guess it is needed to call "chrome.permissions.request" to use optional_permissions.

sindresorhus:
https://github.com/sindresorhus/github-notifier-chrome/blob/f2aec934049bb2f13be20dbb0a6fb005b8865e7b/extension/options.js#L27

bigwheel:
hm... but I didnt see any dialog in Option Page.
I'll try to seek the reason why dialog isnt shown.

sindresorhus:
Chrome seems to barf if you don't end the URL in a `/`. Fixed now.

bigwheel:
Now i can see dialog and it seems working right now.

Thanks!

PR #18, opened 2015-05-02
working with notifications api and oauth token

### LINKED ISSUE #11: Need a way to set update frequency and/or check updates manually
Otherwise it feels too unresponsive.

gcochard:
Fixes #11
Drops need for any special permissions, as github api has proper CORS headers.
Adds requirement for an oauth token generated by the user.

### DIFF
--- a/extension/api.js
+++ b/extension/api.js
@@ -69,6 +69,7 @@
 			callback(new Error('missing token'));
 			return;
 		}
+
 		if (/(^(https:\/\/)?(api\.)?github\.com)/.test(url)) {
 			url = 'https://api.github.com/notifications';
 		} else {
@@ -78,13 +79,14 @@
 
 		xhr('GET', url, opts, function (data, status, response) {
 			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
-			if (status >= 400) {
+
+			if (status >= 500) {
 				callback(new Error('server error'), null, interval);
 				return;
 			}
 
-			if (status === 304) {
-				callback(null, 'cached', interval);
+			if (status >= 400) {
+				callback(new Error('client error: '+data), null, interval);
 				return;
 			}
 


extension/api.js
```diff
@@ -1,23 +1,35 @@
...
 
 	var xhr = (function () {
-		var xhr = new XMLHttpRequest();
-		return function (method, url, callback) {
-			xhr.onreadystatechange = function () {
-				if (xhr.readyState === 4) {
-					callback(xhr.responseText, xhr.status);
+		var xh = new XMLHttpRequest();
+		return function (method, url, headers, callback) {
+			if(!callback && typeof headers === 'function'){
```
kevva: `if (!callback && typeof headers === 'function') {`

extension/api.js
```diff
@@ -1,23 +1,35 @@
...
+			}
+			xh.onreadystatechange = function () {
+				if (xh.readyState === 4) {
+					return callback(xh.responseText, xh.status, xh);
 				}
 			};
-			xhr.open(method, url);
-			xhr.send();
+			xh.open(method, url);
+			if(headers){
```
kevva: ` if(headers) {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
 	window.gitHubNotifCount = function (callback) {
-		var tmp = document.createElement('div');
+		var token = GitHubNotify.settings.get('oauthToken'),
+			opts = {
+				Authorization:'token '+token
+			},
+			participating = GitHubNotify.settings.get('useParticipatingCount') ? '?participating=true' : '',
+			url = GitHubNotify.settings.get('rootUrl');
+
+		if(!token){
```
kevva: `if (!token) {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+			return callback(-4);
+		}
+		if(/github.com\/?/.test(url)){
```
kevva: `if (/github.com\/?/.test(url)) {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+		}
+		url += participating;
 
-		xhr('GET', GitHubNotify.settings.get('notificationUrl'), function (data, status) {
+		/* -- until github supports total count, this will remain commented
+		var lastModifiedIf = GitHubNotify.settings.get('lastModified');
+		if(lastModifiedIf && lastModifiedIf !== 'null'){
+			opts['If-Modified-Since'] = lastModifiedIf;
+		}
+		*/
```
kevva: Remove this and open an issue about it instead.

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+			}
+
+			if (status === 304) {
+				return callback(-3,interval);
 			}
 
-			tmp.innerHTML = data;
+			console.log(data);
+
+			try{
```
kevva: `try {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+				data = JSON.parse(data);
+			} catch(e){
```
kevva: `catch (err) {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+				return callback(-1,interval);
+			}
+
+			console.log(data);
```
kevva: ?

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
-			var participating = (GitHubNotify.settings.get('useParticipatingCount'))
-				? '/participating'
-				: '';
-			var countElem = tmp.querySelector('a[href="/notifications' + participating + '"] .count');
-			if (countElem) {
-				callback(countElem.textContent !== '0' ? countElem.textContent : '');
-			} else {
-				callback(-2);
+			if (data && data.hasOwnProperty('length')) {
+				return callback(String(data.length),interval);
```
kevva: `return callback(String(data.length), interval);`

extension/main.js
```diff
@@ -16,13 +17,31 @@
 	}
 
 	function update() {
-		gitHubNotifCount(function (count) {
+		gitHubNotifCount(function (count, interval) {
+			var period = 1;
+			if(interval !== GitHubNotify.settings.get('interval')){
```
kevva: `if (interval !== GitHubNotify.settings.get('interval')) {`

extension/main.js
```diff
@@ -16,13 +17,31 @@
...
+				GitHubNotify.settings.set('interval',interval);
+				period = interval / 60;
+				if(period < 1){
```
kevva: `if (period < 1) {`

extension/main.js
```diff
@@ -16,13 +17,31 @@
...
+					period = 1;
+				}
+				chrome.alarms.clearAll(function(){
```
kevva: `chrome.alarms.clearAll(function () {`

extension/main.js
```diff
@@ -39,8 +58,12 @@
 	chrome.runtime.onMessage.addListener(update);
 
 	chrome.browserAction.onClicked.addListener(function (tab) {
+		var url = GitHubNotify.settings.get('rootUrl');
+		if(/api.github.com\/$/.test(url)){
+			url = 'https://github.com/';
+		}
 		var notifTab = {
-			url: GitHubNotify.settings.get('notificationUrl') 
+			url: url+'notifications'
```
kevva: `url + 'notifications'`

extension/options.js
```diff
@@ -1,64 +1,71 @@
...
 
 		loadSettings();
 
 		function updateBadge() {
+			GitHubNotify.settings.revert('lastModified');
 			chrome.runtime.sendMessage('update');
 		}
 
+		function normalizeRoot(url) {
+			if(!/^https?:\/\//.test(url)){
```
kevva: `if (!/^https?:\/\//.test(url)) {`

extension/options.js
```diff
@@ -1,64 +1,71 @@
...
+				// assume it is https
+				url = 'https://' + url;
+			}
+			if(!/\/$/.test(url)){
```
kevva: `if (!/\/$/.test(url)) {`

extension/options.js
```diff
@@ -1,64 +1,71 @@
...
+				url += '/';
+			}
+			return url;
+		}
+
+		formRootUrl.addEventListener('change', function () {
+			var url = normalizeRoot(formRootUrl.value);
+			url += 'settings/tokens/new';
```
kevva: Why not just add it directly to the `var url`?

extension/options.js
```diff
@@ -1,64 +1,71 @@
...
-					// TODO: Use a similar message as `successMessage` to show this too
-					console.error('Permission not granted', chrome.runtime.lastError.message);
-				}
-			});
+			updateBadge();
+			loadSettings();
+
+			clearTimeout(successTimeout);
+			successMessage.classList.add('visible');
+			successTimeout = setTimeout(function() {
```
kevva: `successTimeout = setTimeout(function ()`

extension/options.js
```diff
@@ -1,64 +1,71 @@
...
-			chrome.permissions.remove({
-				origins: [GitHubNotify.settings.get('notificationUrl')]
-			});
-
 			GitHubNotify.settings.reset();
 			loadSettings();
 		});
 	});
-})();
+}());
```
kevva: ?

extension/manifest.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "GitHub Notifier",
-	"version": "1.3.3",
+	"version": "1.4.0",
```
kevva: Leave this up to the author.

extension/main.js
```diff
@@ -50,4 +73,4 @@
 	});
 
 	update();
-})();
+}());
```
kevva: ?

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
 			}
+			return callback(-2,interval);
 		});
 	};
-})();
+}());
```
kevva: ?

extension/api.js
```diff
@@ -1,23 +1,35 @@
...
+			xh.open(method, url);
+			if(headers){
+				Object.keys(headers).forEach(function(k){
+					xh.setRequestHeader(k,headers[k]);
+				});
+			}
+			xh.send();
 		};
-	})();
+	}());
```
kevva: ?

extension/api.js
```diff
@@ -1,23 +1,35 @@
+/*global window, localStorage, XMLHttpRequest, GitHubNotify*/
```
kevva: Assume that these weren't here before because of the author didn't want to.

extension/api.js
```diff
@@ -1,23 +1,35 @@
...
-					callback(xhr.responseText, xhr.status);
+		var xh = new XMLHttpRequest();
+		return function (method, url, headers, callback) {
+			if(!callback && typeof headers === 'function'){
+				callback = headers;
+				headers = null;
+			}
+			xh.onreadystatechange = function () {
+				if (xh.readyState === 4) {
+					return callback(xh.responseText, xh.status, xh);
```
kevva:
```js
callback(xh.responseText, xh.status, xh);
return;
```

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+			opts['If-Modified-Since'] = lastModifiedIf;
+		}
+		*/
+
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
-				callback(-1);
-				return;
+				return callback(-1,interval);
```
kevva:
```js
callback(-1, interval);
return;
```

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+			}
+
+			if (status === 304) {
+				return callback(-3,interval);
 			}
 
-			tmp.innerHTML = data;
+			console.log(data);
```
kevva: ?

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
-				callback(-1);
-				return;
+				return callback(-1,interval);
+			}
+
+			if (status === 304) {
+				return callback(-3,interval);
```
kevva:
```js
callback(-3, interval);
return;
```

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
-		var tmp = document.createElement('div');
+		var token = GitHubNotify.settings.get('oauthToken'),
+			opts = {
+				Authorization:'token '+token
+			},
+			participating = GitHubNotify.settings.get('useParticipatingCount') ? '?participating=true' : '',
+			url = GitHubNotify.settings.get('rootUrl');
+
+		if(!token){
+			return callback(-4);
```
kevva:
```js
callback(-4);
return;
```

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+				return callback(-3,interval);
 			}
 
-			tmp.innerHTML = data;
+			console.log(data);
+
+			try{
+				data = JSON.parse(data);
+			} catch(e){
+				return callback(-1,interval);
```
kevva:
```js
callback(-1, interval);
return;
```

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
-				: '';
-			var countElem = tmp.querySelector('a[href="/notifications' + participating + '"] .count');
-			if (countElem) {
-				callback(countElem.textContent !== '0' ? countElem.textContent : '');
-			} else {
-				callback(-2);
+			if (data && data.hasOwnProperty('length')) {
+				return callback(String(data.length),interval);
 			}
+			return callback(-2,interval);
```
kevva:
```js
callback(-2, interval);
return;
```

extension/main.js
```diff
@@ -39,8 +58,12 @@
 	chrome.runtime.onMessage.addListener(update);
 
 	chrome.browserAction.onClicked.addListener(function (tab) {
+		var url = GitHubNotify.settings.get('rootUrl');
+		if(/api.github.com\/$/.test(url)){
```
kevva: `if (/api.github.com\/$/.test(url)) {`

extension/api.js
```diff
@@ -1,23 +1,35 @@
...
+			xh.onreadystatechange = function () {
+				if (xh.readyState === 4) {
+					return callback(xh.responseText, xh.status, xh);
 				}
 			};
-			xhr.open(method, url);
-			xhr.send();
+			xh.open(method, url);
+			if(headers){
+				Object.keys(headers).forEach(function(k){
```
kevva: `Object.keys(headers).forEach(function (k) {`

extension/api.js
```diff
@@ -40,28 +52,61 @@
...
+	}());
 
 	window.gitHubNotifCount = function (callback) {
-		var tmp = document.createElement('div');
+		var token = GitHubNotify.settings.get('oauthToken'),
+			opts = {
+				Authorization:'token '+token
+			},
+			participating = GitHubNotify.settings.get('useParticipatingCount') ? '?participating=true' : '',
+			url = GitHubNotify.settings.get('rootUrl');
```
kevva: Use multiple line `var` statements.

kevva:
One advice for future PRs; try to adhere to the current code style and always assume that the author has written and done stuff for a reason. Makes the job easier for all of us. Less work for you having to correct all the code style mistakes and less work for us having to point them all out before we can actually comment on the functionality.

gcochard:
Thanks, I'll fix these style issues later today.

gcochard:
@kevva I think I have taken care of all style issues. Given that jshint was failing *before* my changes, I had assumed that the `.jshintrc` was out of date and I took the time to update the code with some best practices that I have found helpful. No worries though, I'm happy to stick to established conventions.

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
-			var countElem = tmp.querySelector('a[href="/notifications' + participating + '"] .count');
-			if (countElem) {
-				callback(countElem.textContent !== '0' ? countElem.textContent : '');
-			} else {
-				callback(-2);
+			if (data && data.hasOwnProperty('length')) {
+				callback(String(data.length),interval);
+				return;
 			}
+			callback(-2,interval);
```
arthurvr: `callback(-2, interval);`

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
-			var participating = (GitHubNotify.settings.get('useParticipatingCount'))
-				? '/participating'
-				: '';
-			var countElem = tmp.querySelector('a[href="/notifications' + participating + '"] .count');
-			if (countElem) {
-				callback(countElem.textContent !== '0' ? countElem.textContent : '');
-			} else {
-				callback(-2);
+			if (data && data.hasOwnProperty('length')) {
+				callback(String(data.length),interval);
```
arthurvr: `callback(String(data.length), interval);`

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
+			if (status === 304) {
+				callback(-3,interval);
 				return;
 			}
 
-			tmp.innerHTML = data;
+			try {
+				data = JSON.parse(data);
+			} catch (err) {
+				callback(-1,interval);
```
arthurvr: `callback(-1, interval);`

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
+		} else {
+			url += 'api/v3/notifications';
+		}
+		url += participating;
+
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
-				callback(-1);
+				callback(-1,interval);
```
kevva: `callback(-1, interval);`

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
+				return;
+			}
+
+			if (status === 304) {
+				callback(-3,interval);
```
kevva: `callback(-3, interval);`

extension/api.js
kevva: `callback(-1, interval);`

gcochard:
I added an `.eslintrc` file with rules that match the coding style. Rather than having to review the code and point style issues out individually, you can point potential future contributors to it.

gcochard:
Additionally, I can set up a travis-ci job to automate the check if desired.

attilaolah:
> Additionally, I can set up a travis-ci job to automate the check if desired.
￼ Merge pull request

That would make this a lot easier, I :+1: for that.

attilaolah:
Otherwise great work @gcochard, I'll package this later today and give it a try. Thanks!

gcochard:
@attilaolah Please make sure that this doesn't break existing users. Is there a one-time post-upgrade hook we can use to announce that the notification URL format has changed?

arthurvr:
I'd say discussions about code style tooling shouldn't go in this PR, that's unrelated. Let's stay on topic on this PR and just fix #11.

attilaolah:
Good point, @arthurvr.

@gcochard:

> Is there a one-time post-upgrade hook we can use to announce that the notification URL format has changed?

I don't think so, but then, we could just mention the change before asking for the token.

> On second thought, we'll have to prompt for the OAuth token too, so that's the lesser of two issues.

We need to do this anyway. After all, the token could be revoked as well. If we get an authorisation error from the API, we could:

1. either show a page explaining that we need an authorisation now, and generate a link with a request token, or
2. just generate a link with a request token and redirect there, without much ado.

Furthermore, to avoid confusion, @sindresorhus could create an org for this repo and ask for the token on behalf of the org to make it obvious to users that they are being requested authorisation for this extension, and there's nothing phishy going on in their browser.

eprouty:
It seems like those features of notifying the end user about the token, or automatically redirecting/requesting a token should be a new issue.  This pull solves the problem of using the github api and allowing for a token to be input ala #11.  

Is there something else that needs to be done before this can be merged in?

gcochard:
I'm gonna break this apart into the original token/API stuff, then add separate PRs for eslint and travis-ci.

gcochard:
This is ready for a final review.

attilaolah:
This seems to work in general, however I noticed a few things:

1. The link "Create an oauth token" should read "Creat an OAuth token".
2. The link points to https://{root_host}/settings/tokens/new. Since I'm not using GHE, my root URL is https://api.github.com/, so the link points to https://api.github.com/settings/tokens/new, which is a 404. Even worse, if I leave the custom URL empty, the link will point to https://sennitgs/tokens/new.
3. The above link should also end in `?scopes=notifications` to pre-select the right scope, as a convenience.
4. When I have zero notifications, the button now shows a label containing "0". It used to show no label for when there are no notifications.

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
+			? '?participating=true'
+			: '';
+		var url = GitHubNotify.settings.get('rootUrl');
 
-		xhr('GET', GitHubNotify.settings.get('notificationUrl'), function (data, status) {
+		if (!token) {
+			callback(-4);
+			return;
+		}
+		if (/github.com\/?/.test(url)) {
```
attilaolah: I'd be more restrictive here and put a `[/.]` before `github`. This would catch (unlikely) URLs like `acmegithub.com` or `acme-github.com`.

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
-				callback(-1);
+				callback(-1, interval);
+				return;
+			}
+
+			if (status === 304) {
+				callback(-3, interval);
```
attilaolah: These arguments (-1, -3, -4) look a bit too cryptic, why not name them to make this more readable?

extension/options.html
```diff
@@ -47,9 +47,15 @@
...
-							<p>Specify the notification URL to your GitHub Enterprise</p>
-							<label for="notification_url">Notification URL</label>
-							<input type="text" id="notification_url" autofocus>
+							<p>Specify the root URL to your GitHub Enterprise</p>
+							<label for="root_url">Root URL</label>
+							<input type="text" id="root_url" autofocus placeholder="e.g. https://github.yourco.com/">
+						</section>
+						<section>
+							<h3>OAuth Token</h3>
+							<p><a id="gh_link" href="https://github.com/settings/tokens/new" target="gh_target">Create an oauth token</a> with the <strong>notifications</strong> permission and specify it below</p>
```
attilaolah: Create an OAuth token

extension/main.js
```diff
@@ -16,13 +16,31 @@
 	}
 
 	function update() {
-		gitHubNotifCount(function (count) {
+		gitHubNotifCount(function (count, interval) {
+			var period = 1;
+			if (interval !== GitHubNotify.settings.get('interval')) {
+				GitHubNotify.settings.set('interval', interval);
+				period = Math.ceil(interval / 60);
+				if (period < 1) {
```
attilaolah:
```js
if (period < 1 || isNaN(period)) {
```

Otherwise I get this error:

```
extensions::uncaught_exception_handler:8 Error in response to alarms.clearAll: Error: Invalid value for argument 1. Property 'periodInMinutes': Value must not be NaN.
    at Object.callback (chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/main.js:28:20)
    at chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/main.js:27:19
    at window.gitHubNotifCount (chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/api.js:68:4)
    at update (chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/main.js:19:3)
    at chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/main.js:76:2
    at chrome-extension://nodblbbnmnhfphpaokabkcinaglkedik/main.js:77:3
```

extension/main.js
```diff
@@ -16,13 +16,31 @@
...
-					text = 'Unable to find count on page';
+					text = 'Unable to find count';
+				} else if (count === -3) {
+					// not-modified, do not re-render
+					return;
+				} else if (count === -4) {
+					text = 'Missing access token, please create one and enter it in Options';
 				}
 				render('?', [166, 41, 41, 255], text);
 			} else {
```
attilaolah: At this point, `typeof(count) === "string"`, so the next `if` will not work.

extension/main.js
```diff
@@ -16,13 +16,31 @@
...
+				if (period < 1) {
+					period = 1;
+				}
+				chrome.alarms.clearAll(function () {
+					chrome.alarms.create({periodInMinutes: period});
+					chrome.alarms.onAlarm.addListener(update);
+					chrome.runtime.onMessage.addListener(update);
+				});
+			}
 			if (count < 0) {
```
attilaolah: Above this line, add:

```js
if (typeof(count) === "string") {
    count = parseInt(count, 10);
}
```

attilaolah: Scratch that. In fact, all you need to do is get rid of the `String()` that wraps `data.length` in `api.js`.

extension/api.js
```diff
@@ -43,25 +55,51 @@
...
-			var participating = (GitHubNotify.settings.get('useParticipatingCount'))
-				? '/participating'
-				: '';
-			var countElem = tmp.querySelector('a[href="/notifications' + participating + '"] .count');
-			if (countElem) {
-				callback(countElem.textContent !== '0' ? countElem.textContent : '');
-			} else {
-				callback(-2);
+			if (data && data.hasOwnProperty('length')) {
+				callback(String(data.length), interval);
```
attilaolah: No need for the `String()` here, the first parameter should be an integer, not a string.

extension/main.js
attilaolah: Actually, scratch the above comment as well.

The root of the problem is that `GitHubNotify.settings.get('interval')` is a string, so it will never `=== interval`. Either convert that to a number, or make sure the value you get from settings is always a number.

attilaolah: I would probably add something like this to `GitHubNotify.settings`:

```js
getNumber: function (name) {
    return parseInt(this.get(name), 10);
}
```

and call that to avoid having to `parseInt(…)` every time.

extension/main.js
attilaolah: Scratch this one as well. See my below comment why this is a string in the first place.

attilaolah:
To summarise all my above comments, @gcochard, here's what I would change:

```diff
diff --git a/extension/api.js b/extension/api.js
index 2752369..3571d49 100644
--- a/extension/api.js
+++ b/extension/api.js
@@ -95,7 +95,7 @@
 			}
 
 			if (data && data.hasOwnProperty('length')) {
-				callback(String(data.length), interval);
+				callback(data.length, interval);
 				return;
 			}
 			callback(-2, interval);
diff --git a/extension/main.js b/extension/main.js
index b067bb4..d379411 100644
--- a/extension/main.js
+++ b/extension/main.js
@@ -18,7 +18,7 @@
 	function update() {
 		gitHubNotifCount(function (count, interval) {
 			var period = 1;
-			if (interval !== GitHubNotify.settings.get('interval')) {
+			if (interval !== parseInt(GitHubNotify.settings.get('interval'), 10)) {
 				GitHubNotify.settings.set('interval', interval);
 				period = Math.ceil(interval / 60);
 				if (period < 1) {
diff --git a/extension/options.html b/extension/options.html
index 96bac1f..f3048e3 100644
--- a/extension/options.html
+++ b/extension/options.html
@@ -53,7 +53,7 @@
 						</section>
 						<section>
 							<h3>OAuth Token</h3>
-							<p><a id="gh_link" href="https://github.com/settings/tokens/new" target="gh_target">Create an oauth token</a> with the <strong>notifications</strong> permission and specify it below</p>
+							<p><a id="gh_link" href="https://github.com/settings/tokens/new" target="gh_target">Create an OAuth token</a> with the <strong>notifications</strong> permission and specify it below</p>
 							<label for="oauth_token">OAuth Token</label>
 							<input type="text" id="oauth_token">
 						</section>
```

attilaolah:
I also sent a PR here: https://github.com/gcochard/github-notifier-chrome/pull/1, to make life simpler.

attilaolah:
I also just noticed that clicking on the button doesn't update the notification count, I have to wait for the one minute interval to pass.

attilaolah:
Actually, nevermind that. That is a separate issue, and we can tackle in a separate PR. We'll need some kind of polling, which we can do once this PR is merged.

gcochard:
@attilaolah Thanks! It's all merged in.

attilaolah:
OK, in general, I'm :+1: for merging this as it stands now.

It is a little annoying that the counter no longer updates in real time, but I'd say we merge this, then start working on a fix for that, before this update gets released to the Chrome marketplace.

Is everyone fine with this, can we push <kbd>the button</kbd>?

extension/api.js
eprouty: This should be `/\.github\.com\/?/.test(url)` otherwise the current `.` act as regex wildcard characters.

eprouty:
Not quite ready for a merge yet, I think the above regex for checking `github.com` needs to be updated.  

Changing the cryptic return numbers to something named would be a nice addition as well.

attilaolah:
Yeah, :+1: for the regex, good catch. Let's fix these two issues then, @gcochard can you make [the changes](#issuecomment-102055219) mentioned by @eprouty, please? :bowtie:

attilaolah:
I just noticed an issue:

For some reason, my counter is now stuck at "1". Looking at the network tab in developer tools, I see that the polling returns 304s. It would appear that I somehow managed to clear my notifications in an interval that is earlier than what is supplied in the "If-Modified-Since" header, so I never get the update.

I'm not sure how did this happen, but it is pretty annoying.

attilaolah:
In fact, even reloading the extension didn't help. I had to disable cache in the developer tools (network tab), then reload the extension. Now everything is back to normal, however, requests no longer use the `If-Modified-Since` header. They now include an `If-None-Match`, with the `ETag` received from the previous request. Also, interestingly, the previous (initial) request now doesn't contain a `Last-Modified` header.

gcochard:
Really weird. I *had* the `If-Modified-Since` and `If-None-Match` headers at one point as an experiment, but it [didn't work](https://github.com/sindresorhus/github-notifier-chrome/issues/11#issuecomment-98216511).

The `If-None-Match` is probably added by the browser; regardless, the response's `ETag` header is set to `nil` so it should always get a fresh response.

attilaolah:
OK, I can reproduce the issue now:

1. Inspect the extension's background page, open the network tab.
2. Create an issue, and wait for the notification poll to receive it. In my case, I got a 200, with `Last-Modified: Thu, 14 May 2015 15:43:10 GMT`.
3. Clear the issue on GitHub, and wait for the next request.
4. The next request comes, but it doesn't even make it to the server: it gets cached by the client ("from cache" in network tab). In JS, it appears as a 200 with a `Last-Modified` header from point 2. above.

Subsequent requests alternate between "from cache" and 304, however, the 304 is interpreted as the response containing the issue, and does not update the badge.

attilaolah:
Note that the GitHub API conveniently includes a `Status` header in the response, so it is quite obvious when something comes from the browser's cache: the response status is "200 OK", but the response contains a `Status` header with the value `304 Not Modified` (in the case where the previous request was a 304).

gcochard:
Ugh, we're gonna have to include a cachebuster.

attilaolah:
I hope we don't, that would be really ugly.

attilaolah:
In fact, I'm not even sure anymore whether this is a bug in the GitHub API or in this extension.

The request that gets sent contains an `If-Modified-Since: Thu, 14 May 2015 15:43:10 GMT` header, the same timestamp that I got with the request containing a 200 OK (both requests actually hit the server, I double-checked that).

So the browser, correctly, assumes that the content didn't change, and fills in the payload containing a single notification, and the npasses it on to the JS. When I make the same request with `curl`, I get a 304, as expected. But when I make the same request without the `If-Modified-Since` header, I get an empty array.

I'll have another look at the API docs or get in touch with GitHub support to see why this is happening.

attilaolah:
Also, inspecting the `xhr` response, there seems to be no way to figure out whether this was a 200 or a 304. I see 304 responses in the network tab, but setting a breakpoint in JS tells me it is 200. The browser makes the cache fully transparent, I don't think there's much we can do about this.

attilaolah:
Hold on, I have an idea:

It seems that the response sets `Cache-Control: private, max-age=60, s-maxage=60`. However, our interval is exactly 60, because that is what we get from `X-Poll-Interval`. Let me try adding an extra second, to make sure we don't make a request during the time that we know the browser will serve us a cached copy anyway.

attilaolah:
c1d63e6 seems fine, although for me this also seems to fix the problem:

```diff
diff --git a/extension/api.js b/extension/api.js
index 3571d49..7b46c1d 100644
--- a/extension/api.js
+++ b/extension/api.js
@@ -77,6 +77,16 @@
 
                xhr('GET', url, opts, function (data, status, response) {
                        var interval = Number(response.getResponseHeader('X-Poll-Interval'));
+                       var cacheControl = response.getResponseHeader('Cache-Control');
+                       if (cacheControl) {
+                               var match = cacheControl.match(/\bmax-age=(\d+)/);
+                               if (match) {
+                                       var maxAge = parseInt(match[1]);
+                                       if maxAge >= interval {
+                                               interval = maxAge + 1;
+                                       }
+                               }
+                       }
                        if (status >= 400) {
                                callback(-1, interval);
                                return;
```

attilaolah:
Actually, no, it doesn't. Let me try c1d63e6.

gcochard:
I think I want to move to `when` instead of `intervalInMinutes` for the alarms, just to make it more explicit and give us *at least* `interval` seconds between polling, and to give us time resolution > 1 minute.

attilaolah:
Yeah, you're right. Also, this way you can more easily change the interval while running.

Another thing with the interval is that it probably hit the cache because it was strictly one minute, however, it takes some time to make the actual request, so the time between two requests is slightly less than the interval, which is why every other request was from the local cache.

extension/main.js
```diff
@@ -16,31 +16,51 @@
...
-		gitHubNotifCount(function (count) {
+		gitHubNotifCount(function (count, interval) {
+			var period = 1;
+			if (interval !== parseInt(GitHubNotify.settings.get('interval'), 10)) {
+				GitHubNotify.settings.set('interval', interval);
+				period = Math.ceil(interval / 60);
+				if (period < 1) {
+					period = 1;
+				}
+                chrome.alarms.create({when: Date.now() + 2000 + (period * 60 * 1000)});
```
attilaolah: You shouldn't `* 60` this, `period` is already in seconds.

attilaolah: Actually, I take that back. I just noticed the `/60` above.

However, for some reason, my extension doesn't poll every minute now, and I don't know why.

attilaolah: Aah, got it: you are only creating another alarm if `interval !== parseInt(…)`. You need to move thin down one line.

attilaolah: OK I tried moving this down by one line, and that fixes the weird browser-cache issue, now each request hits the server.

However, I'm still getting strange results, whereby the counter is still stuck at 1, even though I have no unread notifications. I get 304s from the API (and they don't even affect the rate limit — hooray for that!), but the browser backfills the request with what seems to be bad content.

attilaolah:
The only explanation I can find for this is that the API doesn't work correctly. It is giving me 304s for `If-Modified-Since: Thu, 14 May 2015 16:37:54 GMT`, which is the timestamp from the `Last-Modified` header of the last request that contained a single notification, even though I have no notifications now.

I'll get in touch with GitHub support.

attilaolah:
I sent a support request, maybe the GitHub devs can help with this.

gcochard:
I already contacted GH support, it only includes *new* notifications as a modification and is unlikely to change until API v4.

>> Currently trying to get an unread notification count using conditional requests. It turns out that the `If-Modified-Since` header is only valid when *new* notifications have been added, and ignores when notifications have been marked as read. This causes problems with using the API for the GitHub Notifier Extension, as shown here: https://github.com/sindresorhus/github-notifier-chrome/issues/11#issuecomment-98216511
>>
>> Is there a way to change that behavior, or better yet, add a method for `/notifications/count` that respects the `If-Modified-Since` header?

> Thanks for reaching out. Changing the behavior of that endpoint might break existing clients which are relying on the current behavior. For that reason, the semantics of that endpoint will probably not be changing in this major version of the API (v3), but I'll mention your feedback to the team so that they can consider it for the next major version (v4).

> As for adding a new endpoint -- I'll mention that to the team as well, but I can't make any promises (if I had to guess, I'd say that it won't happen in the near future). I do agree such an endpoint would be helpful, though.

gcochard:
Apparently [chrome will **always** transparently cache files unless `no-store` is set](https://code.google.com/p/chromium/issues/detail?id=28035#c5). It will check for modification with the `If-Modified-Since` set to the `Last-Modified` value from the last response, so we are hitting the same issue, albeit transparently rather than explicitly. AFAIK there is no way to turn this off.

We can work around the issue by explicitly setting the `If-Modified-Since` header to `''`. I will include this in the next push.

attilaolah:
OK, thanks @gcochard.

For the record, the current behaviour breaks the HTTP spec (*any* change to the resource, even marking a notification as read, that modifies the payload, counts as modifying the resource, and a 304 should not be used — although RFC2616 isn't too strict about this). Pity because all other parts of the API obey HTTP so nicely.

I also understand that this is not going to change easily, since there might be clients out there depending on this behaviour.

In any case, the end result is that our badge is no longer working because the browser caches the request. So we either include a cache buster (which I hoped to avoid), and affect the rate limit this time for real (the limit is high enough anyway), or we break this extension for existing users, which also sucks.

gcochard:
Let me know if the last push works for you, it worked for me since setting the header to `''`.

attilaolah:
Thanks, a93a2af works for me so far, I no longer get any cached responses. See my note about the commit though — I'm checking now whether the `ETag` based cache would be triggered. So far it seems not.

So I'm good with this. How about we give the negative counts some variable names, as @eprouty suggests, and this would be good to merge then?

attilaolah:
This just hit me — if we wanted to be a bit nicer on the rate limit, we could only enable the cachebuster when the last count that we received is non-zero. If if it zero, the API will work according to the HTTP spec.

Let me try that.

gcochard:
Definitely. I will add those vars in the morning (US PDT), and add the logic for only cache busting on non-zero count.

attilaolah:
Yeah, I just noticed minutes ago that you're in PDT, go get some :sleeping:, I'll make the changes and test this throughout the day. If that works I'll send another PR :bowtie:

gcochard:
@attilaolah Try it out now, should be good to go! :shipit:

attilaolah:
Unfortunately this doesn't work now: it keeps alternating between zero and nonzero count for every request.

Here's why: When the cachebuster is enabled, I get a 200, with zero count — that's fine. But because the count is now zero, the next request will not include the cachebuster, so I'll get a 304 — still fine. However, Chrome back-fills the request with a payload containing a single notification. Which sets the `lastCount` to 1, so the next request will have a cachebuster, and the the count goes back to zero, and so on.

It seems a little like Chrome should be smart enough to use the last received response for the cache, but instead it seems to use the oldest one after the `Last-Modified` header?

In any case, this won't work it seems. I'll try to clear the cache just in case, but I doubt that would solve the problem.

gcochard:
It seems like github's api is not sending back a `Last-Modified` header in certain cases. Should we keep track of the last one, and send it if we have it, otherwise send `''`?

attilaolah:
> Should we keep track of the last one, and send it if we have it, otherwise send `''`?

Hmm, sounds like an idea. Let's try that.

gcochard:
Just tried that, it looks like chrome is giving us the old, stale header from the stale response. FWIW mine is alternating between 0 and 16 notifications, with a stale response from the 14th.

attilaolah:
Looks like we're stuck with an unconditional cachebuster until this is resolved then.

gcochard:
Reverted. Good to go?

attilaolah:
I'd probably just `git reset --hard ef6322c` this instead of keeping revert branches, but I don't really mind this way either.

I think this is good to go now. Everyone, :shipit:?

gcochard:
I can always squash the whole PR into one commit.

attilaolah:
> I can always squash the whole PR into one commit.

Nah, I don't think that's necessary. But like I said, I don't mind. Whichever way is easier for you.

Important thing is that this works now, finally.

gcochard:
Ok, I just reset to ef6322c.

attilaolah:
I was just thinking. If there was a way to get the `Status` or `Date` header from the response, we could figure out whether this was a 200 or a 304 response (iow bypass Chrome's transparent caching), and use that with the conditional cachebuster when the current count is zero, ignoring what Chrome might have put in the response payload.

We can't do this because these headers are not exposed by CORS. Does anyone know if there's a way to bypass CORS from extensions, like using a permission or something?

---

I still think we should merge this, regardless of my above comment. Trying to preserve the rate limit is a separate issue.

extension/api.js
```diff
@@ -43,25 +56,51 @@
...
+
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
-				callback(-1);
+				callback(new Error('server error'), null, interval);
+				return;
+			}
+
+			if (status === 304) {
```
attilaolah: It seems to me like this branch will never be executed.

extension/api.js
```diff
@@ -43,25 +56,51 @@
...
+		if (/(^(https:\/\/)?(api\.)?github\.com)/.test(url)) {
+			url = 'https://api.github.com/notifications';
+		} else {
+			url += 'api/v3/notifications';
+		}
+		url += participating;
+
+		xhr('GET', url, opts, function (data, status, response) {
+			var interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			if (status >= 400) {
```
attilaolah: Note that a bad token would probably result in a 401, and it is technically not a server error. Maybe a `>= 500` would be more acceptable here, and a separate check for `>= 400` or even one for `== 401`.

gcochard:
Both of your latest comments addressed.

attilaolah:
Awesome. I really don't see anything else to improve here, I'd say we :shipit:.

eprouty:
Agreed, :shipit:

gcochard:
:shipit: :rocket: :package:

sindresorhus:
Thanks all. Excellent work :)

:ship: d! Should be live in the Chrome Webstore in an hour or so.

kevva:
Sweet.

PR #19, opened 2015-05-14
Simplify GitHubNotify.settings

attilaolah:
Simplify `GitHubmNotify.settings`. I don't see why we couldn't use `localStorage.clear` here.

### DIFF
--- a/extension/api.js
+++ b/extension/api.js
@@ -32,10 +32,7 @@
 					return item;
 				},
 				set: localStorage.setItem.bind(localStorage),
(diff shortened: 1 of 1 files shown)


arthurvr:
Yeah, makes sense :+1:.

attilaolah:
Hold on for a sec. There's another bug.

attilaolah:
OK, since nobody objects, I will push the <kbd>button</kbd>. After all we can always revert.

PR #23, opened 2015-06-01
Hotfix in case of empty/invalid access token

ArnaudLigny:


extension/main.js
```diff
@@ -29,24 +29,32 @@
 				}
 			}
 
+			if (interval !== 'undefined') {
+				period = 1;
```
eprouty: This is overwriting what was set above...  in every case except when `interval === 'undefined'` you overwrote what was determined for `period`.  What is the intention here?

Narno: Hum, it's a big mistake: The test should be ```if (interval === 'undefined')```. It's a dirty test to avoid case of the auth token is empty.

Narno:
In any case, that PR is not intended to be merge, but to highlight some problems with the access token :-)

extension/main.js
```diff
@@ -29,24 +29,32 @@
 				}
 			}
 
+			if (interval !== 'undefined') {
```
kevva: `if (interval) {`

eprouty:
Alright, you should add a description indicating that next time.  Hotfix implies that it is meant to be merged asap after all.

A better way to do this would be to add a comment to the issue tracking this problem.  Then the dev who is going to implement the merge-able solution can reference this suggestion easier.

Narno:
OK Eric, sorry for that!

eprouty:
No problem, open source is all a learning process... I'm very new to this project as well so maybe someone will come back and tell me I'm wrong about this :)

Look forward to seeing your comments in the issue though, glad you are helping out with the project! Thanks @Narno !

sindresorhus:
od3vvoq

PR #24, opened 2015-07-03
Add CI, ESLint, and Travis

gcochard:
- Adding travis config
- Add pre-commit hook to lint and pre-push to test
- Add mocha tests
- Move to window.<name> everywhere.
  - This is necessary for tests to work properly, and does not cause any
    regressions in the browser. Tests currently use mocks for Chrome.*
    methods and localstorage.

PR #25, opened 2015-07-07
Desktop notifications (fixes #12)

### LINKED ISSUE #12: Use Notifications API to get desktop notifications
Hi! Great extension, I love it!

I currently have email notifications activated for _everything_ on GitHub, since I can't miss out on stuff. The notifications icon in this extension is a bit too unobtrusive – would it be possible to support opt-in desktop notifications for GitHub?

Thanks!

inikulin:
Fixes #12.

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -13,6 +13,15 @@
 		chrome.browserAction.setTitle({
 			title: title
 		});
+
+		if (badge && window.GitHubNotify.settings.get('showDesktopNotif')) {
+			chrome.notifications.create('github-notifier', {
+				title: 'GitHub',
+				iconUrl: 'icon-128.png',
+				type: 'basic',
+				message: 'You have new unread notifications'
+			}, function () { });
+		}
 	}
 
 	function update() {
@@ -67,12 +76,8 @@
 		});
 	}
 
-	chrome.alarms.create({when: Date.now() + 2000});
-	chrome.alarms.onAlarm.addListener(update);
-	chrome.runtime.onMessage.addListener(update);
-
-	chrome.browserAction.onClicked.addListener(function (tab) {
-		var url = window.GitHubNotify.settings.get('rootUrl');
+	function openNotifPage(tab) {
+		var url = GitHubNotify.settings.get('rootUrl');
 
 		if (/api.github.com\/$/.test(url)) {
 			url = 'https://github.com/';
@@ -82,11 +87,19 @@
 			url: url + 'notifications'
 		};
 
-		if (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === notifTab.url) {
+		if (tab && (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === notifTab.url)) {
 			chrome.tabs.update(null, notifTab);
 		} else {
 			chrome.tabs.create(notifTab);
 		}
+	}
+
+	chrome.alarms.create({when: Date.now() + 2000});
+	chrome.alarms.onAlarm.addListener(update);
+	chrome.runtime.onMessage.addListener(update);
+	chrome.browserAction.onClicked.addListener(openNotifPage);
+	chrome.notifications.onClicked.addListener(function () {
+		openNotifPage(null);
 	});
 
 	update();
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -12,7 +12,8 @@
 	},
 	"permissions": [
 		"alarms",
-		"activeTab"
+		"activeTab",
+	  	"notifications"
 	],
 	"background": {
 		"persistent": false,
--- a/extension/options.html
+++ b/extension/options.html
@@ -62,6 +62,11 @@
 							<input type="checkbox" id="use_participating">
 							<label for="use_participating">Only show unread count for issues I'm participating in</label>
 						</section>
+						<section>
+							<h3>Desktop Notifications</h3>
+							<input type="checkbox" id="show_desktop_notif">
+							<label for="show_desktop_notif">Show desktop notifications</label>
+						</section>
 						<div class="controls">
 							<button id="save">Save</button>
 							<button id="reset">Reset to Default</button>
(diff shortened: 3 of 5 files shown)


inikulin:
Arghh, requires merge of the #24 and rebase :tired_face:

inikulin:
\cc @sindresorhus

extension/main.js
```diff
@@ -1,6 +1,11 @@
 (function () {
 	'use strict';
 
+	function createNotificationId() {
+		var id = Math.floor(Math.random() * 9007199254740992) + 1;
```
Narno: Why create a randomized ID? What is the goal?

Narno:
:+1:

extension/main.js
inikulin: It's not clear from the docs if notification ID should be unique. Therefore I just took this snippet from the example in docs.

Narno: The notification ID should be unique if you want to clear the previous notification before.
In the example given by Google (https://github.com/GoogleChrome/chrome-app-samples/tree/master/samples/gcm-notifications), a new notification is created for each new message receive.
In our case, there is only one "source" of notification : The user account. I think it's not necesseray to create a new ID each time. On the contrary.

attilaolah: > On the contrary.

Agreed. Limiting ourselves to a single notification seems like the desirable behaviour.

inikulin:
Fixed

sindresorhus:
// @gcochard @attilaolah @LeoColomb @chrisbarmonde @jamime

chrisbarmonde:
:+1:

LeoColomb:
:+1:, works well.

LeoColomb:
Oh, except in the setting view, where value is not loaded: https://github.com/inikulin/github-notifier-chrome/commit/bda8a65d9e0393e1d80991f5fca268806e0627b5#diff-2f54bc484294ab8621cfcfae3033c23fR16

sindresorhus:
@inikulin #24 was merged. Mind rebasing? :)

inikulin:
Rebased, fixed value loading in settings and tests.

inikulin:
ping

extension/main.js
```diff
@@ -13,6 +13,15 @@
 		chrome.browserAction.setTitle({
 			title: title
 		});
+
+		if (badge && window.GitHubNotify.settings.get('showDesktopNotif')) {
+			chrome.notifications.create('github-notifier', {
+				title: 'GitHub',
```
gcochard: Maybe change the title to `GitHub Notifier` rather than simply `GitHub`. /cc @sindresorhus what do you think?

sindresorhus: :+1:

extension/main.js
```diff
@@ -13,6 +13,15 @@
...
+				iconUrl: 'icon-128.png',
+				type: 'basic',
+				message: 'You have new unread notifications'
```
sindresorhus: I wonder how much more useful this is than just the badge. I was thinking with notifications we could show the title of the actual GitHub notification.

sindresorhus: We could then use the `buttons` to have a button for `View` and one for `Mark as read`.

inikulin: @sindresorhus This is quite useful when you have Chrome minimized or as background window, e.g. while working on code in IDE. `Mark as read` button doesn't makes much sense, since in most cases you need to take a look on notifications before marking them as read.

attilaolah: A <kbd>View</kbd> button would be redundant since we could ideally make clicking on the notification itself open the issue in question.

That said, I fully agree with @sindresorhus in that it would be more useful to show the actual notifications themselves. I'd go with one notification per issue.

attilaolah: A <kbd>Mark as read</kbd> also makes sense for short notifications. If it's someone merging your PR, there's no real value in opening it since you already have all the information, so marking it as read saves time.

inikulin: I'm not sure I would like to have a "notifications viewer" instead of notifier. The only goal of this plugin is to be notified ASAP. I prefer GH interface to read/deal with the notifications.

sindresorhus: Ok, let's ignore the buttons. Though I still feel it would be much more useful if you could see the notification title. Then I would know if it's something I care about or not, instead of just being notified for random unimportant notifications.

extension/main.js
```diff
@@ -13,6 +13,15 @@
 		chrome.browserAction.setTitle({
 			title: title
 		});
+
+		if (badge && window.GitHubNotify.settings.get('showDesktopNotif')) {
```
Narno: I think it should also check if the number of notifications is higher than the previous audit: without this test, the desktop notification window is shown again and again, as long we did not read all notifications. It is painful! :-/

extension/main.js
kayone: @inikulin I'm really excited about this PR getting merged in, however I think I agree with @sindresorhus that it would make it a lot more useful if you could include the title or at least the repository the notification is from. I don't think the argument is about if this should be a viewer or notifier it's about giving the user enough info so they can make a decision on whether or not go to github to view the message. I get tons of notifications from different repos and it would help me a lot if I could tell where the notification is from, if I'm at work and it's a work repo then I should probably check it, but if it's from one of my hobby projects it's ok to ignore.

TheBox193: I say most of these items can be enhancements to make notifications better that can be introduced in future PRs.

What items are critical to be added in?

If this base version of notifications is good, let's open issue `feature requests` issues for them to track.

extension/main.js
Narno: @inikulin It should be easier now because we set the current "count" value in storage: https://github.com/sindresorhus/github-notifier-chrome/blob/master/extension/main.js#L57

YuriSolovyov:
Is this PR going to be updated? Do we still want it? Can anyone try to catch-up?

sindresorhus:
Yes, if someone resolves the review feedback.

YuriSolovyov:
If I got it right, main feedback is that notification should show title of GitHub issue and clicking on it  should get you at issue/pr page?

YuriSolovyov:
ATM I got something like this:
image

PR #27, opened 2015-07-20
add version info in panel

PeterDaveHello:
So we don't need to open store to know the current version!

### DIFF
--- a/extension/options.html
+++ b/extension/options.html
@@ -23,6 +23,10 @@
 			#success_message.visible {
 				opacity: 1;
 			}
+			
+			#id {
+				float: right;
+			}
 		</style>
 		<script src="api.js"></script>
 		<script src="options.js"></script>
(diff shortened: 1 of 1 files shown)


arthurvr:
Isn't the store simply a better place to keep track of this? That's where everybody will logically be searching, and we don't have to update this version every time.

PeterDaveHello:
hmmmm ... I think this can be local, so no need to open the website, just like this project does - get notifications but don't need to open the website, faster and easier if I just wanna know the version. But this is just my point :smile:

extension/options.html
```diff
@@ -23,6 +23,10 @@
 			#success_message.visible {
 				opacity: 1;
 			}
+			
+			#id {
```
attilaolah: The `id` of the `<p>` is `ver`, not `id`.

attilaolah:
> and we don't have to update this version every time

:+1: 

Plus, I don't think knowing the version should be that important to the end user.

Plus, you don't really need to go to the store to see the version, just visit chrome://extensions/ or navigate from the menu.

PeterDaveHello:
Oh yeah, that's true!

arthurvr:
Agreed.

PR #29, opened 2015-07-20
Chrome bootstrap

PeterDaveHello:


arthurvr:
Does it make much sense to use the minified version? The bytes don't have to be sent over the network so it doesn't matter too much.

PeterDaveHello:
Yeah, may not be too much, but I think it won't bite :smile:

attilaolah:
I actually don't mind using the minified version. In theory, that should result in slightly lower memory usage and a few optimisations as an extra bonus.

That said, I don't think it matters too much. But if I had to cut a tie, I'd go with the minified version as well. Just my 2¢.

sindresorhus:
Thanks. I cherry-picked the first commit: https://github.com/sindresorhus/github-notifier-chrome/commit/50b37efd161924f435b4ee66e3abc8608f8a12a6

> In theory, that should result in slightly lower memory usage and a few optimisations as an extra bonus.

The CSS is parsed into pretty much the same AST in-memory regardless of minification or not. It might save some bytes, but that's neglible. The benefit of un-minified is easier debugging.

Pretty much:

> "In theory there is no difference between theory and practice. In practice there is."

attilaolah:
Oh shucks, I just quickly skimmed through this and thought this was a `.js` file, in which case I would still prefer the minified version.

As for CSS, I fully agree, there's no need for the minified version there.

PR #34, opened 2015-09-12
Bug fix. Sometimes the setting of 'period' could contain a string.

TheBox193:
Doing a quick check to prevent the subsequent math to not fail / evaluate to NaN.

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -17,16 +17,17 @@
 
 	function update() {
 		window.gitHubNotifCount(function (err, count, interval) {
+			var intervalSetting = parseInt(window.GitHubNotify.settings.get('interval'), 10);
 			var period = 1;
 			var text;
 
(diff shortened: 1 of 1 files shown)


extension/main.js
```diff
@@ -23,10 +23,10 @@
 			if (interval !== null && interval !== parseInt(window.GitHubNotify.settings.get('interval'), 10)) {
 				window.GitHubNotify.settings.set('interval', interval);
```
gcochard: Seems like you should check for `interval` not being a number here (or even before this if block). I'd then set `interval` to `60` if it's not a number. This would keep `interval` from getting saved as an invalid value.

TheBox193: Hmm good point.
I noticed that `window.GitHubNotify.settings.get('interval')` can return a string that can't be parsed as an integer. Most notably when a fresh install it can return `"undefined"`. It would be good to be defensive about this somehow.

I'll put in a commit that address that more closely.

sindresorhus:
Thanks :)

PR #35, opened 2015-09-12
Link to participating notifications if only notifications set for participating in settings.

TheBox193:


extension/main.js
```diff
@@ -86,7 +86,11 @@
 			ghTab.url = url + 'notifications';
 		}
 
-		if (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === ghTab.url) {
+		if (window.GitHubNotify.settings.get('useParticipatingCount')) {
+			ghTab.url += '/participating';
+		}
+
+		if (typeof tab !== 'undefined' && (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === ghTab.url)) {
```
sindresorhus: Why `typeof tab !== 'undefined'`? I don't think it can ever be undefined?

TheBox193: In the case of dev console I believe I ran into tab not being set.
I can remove if you wish. I've found it generally a good idea to test the existence of each level of an object before accessing subkeys.

Shall I remove it?

sindresorhus: I guess you can leave it, but the docs say a tab is to be returned, so that's either a doc bug or behaviour bug. https://developer.chrome.com/extensions/browserAction#event-onClicked

TheBox193: Yeah. Alright thanks.

PR #40, opened 2015-09-18
Auto-save options

### LINKED ISSUE #38: Auto-save options
A "Save" button is so 2000. We can instead just auto-save when the user changes the settings by listening to the `change` event on the form inputs. This way we can also remove the alert message thingy.

Narno:
- listens change event on each form imput
- removes controls (« save » and « reset »)
- removes the alert message

Fixes #38.

### DIFF
--- a/extension/options.js
+++ b/extension/options.js
@@ -3,11 +3,9 @@
 
 	document.addEventListener('DOMContentLoaded', function () {
 		var formRootUrl = document.getElementById('root_url');
-		var ghSettingsUrl = document.getElementById('gh_link');
 		var formOauthToken = document.getElementById('oauth_token');
 		var formUseParticipating = document.getElementById('use_participating');
-		var successMessage = document.getElementById('success_message');
-		var successTimeout = null;
+		var ghSettingsUrl = document.getElementById('gh_link');
 
 		function loadSettings() {
 			formRootUrl.value = GitHubNotify.settings.get('rootUrl');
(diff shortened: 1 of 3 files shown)


sindresorhus:
Nice! Looks pretty good. One thing I noticed though, is that when you remove the URL it default back to `https://`. It should rather default to the default GitHub API URL.

extension/options.js
```diff
@@ -33,35 +31,28 @@
...
-		});
-
-		formUseParticipating.addEventListener('change', function () {
-			GitHubNotify.settings.set('useParticipatingCount', formUseParticipating.checked);
+			var url = normalizeRoot(formRootUrl.value);
+			var urlSettings = normalizeRoot(formRootUrl.value) + 'settings/tokens/new?scopes=notifications';
+			// case of url is empty: set to default
+			if (url === normalizeRoot('')) {
+				GitHubNotify.settings.remove('rootUrl');
+				url = GitHubNotify.settings.get('rootUrl');
```
sindresorhus: Why are you trying to get the `rootUrl` right after removing it?

sindresorhus:
```js
var defaultUrl = normalizeRoot('');
if (url === defaultUrl) {
  GitHubNotify.settings.remove('rootUrl');
  url = defaultUrl;
}
```

?

Narno: Because the only way to fetch the default value is the "GitHubNotify.settings.get" method. No?

sindresorhus: Ah, yeah. Makes sense.

sindresorhus:
Looks good. Thanks :)

Issue #39, opened 2015-09-15
Use `chrome.storage.sync` instead of `localStorage`


sindresorhus:
It syncs and supports more types than just strings, so we can simplify the code a bit.

Example usage: https://developer.chrome.com/extensions/optionsV2#step-2
Docs: https://developer.chrome.com/extensions/storage

Narno:
"storage.sync" or "storage.local"? :)

https://developer.chrome.com/extensions/storage#using-sync

sindresorhus:
`storage.sync`

Narno:
I thinks we should let user choose the storage area (local or "sync").

Narno:
WIP, please checks the [dedicated branch](https://github.com/sindresorhus/github-notifier-chrome/tree/feature-39-chrome.storage.sync).

sindresorhus:
Do a PR ;)

Narno:
I should fixes some problems before, introduced by the asynchronous logic of chrome storage methods.

sindresorhus:
PRs doesn't have to be finished. Just easier to review than looking at a fork branch.

Narno:
Yes, it sound good. I will do that today.

YurySolovyov:
I'm thinking of picking that up.
There is a question about migrating local storage values: do we really want it? I noticed a to-do in @Narno PR, but it was not finished.

sindresorhus:
@YurySolovyov Yes, should be easy: https://github.com/sindresorhus/notifier-for-github-chrome/pull/42#issuecomment-142986917

Narno:
> I'm thinking of picking that up.

\o/

Be careful: my code is very old and is decorrelated from the main branch.

YurySolovyov:
@Narno notice that I'm going to pick up an issue, not a PR :)

YurySolovyov:
Note to self:
>```
Drop storage lookup in PermissionsService and only use
chrome.permissions.contains
to check permissions
```

YurySolovyov:
This should be much easier with `async / await` shipped in Chrome 55

Narno:
And we can consider my previous work totally outdated! :-)

sindresorhus:
I think we can make it even simpler by moving to a good module for this, see https://github.com/sindresorhus/notifier-for-github-chrome/issues/147

PR #42, opened 2015-09-24
Switch from `localStorage` to `chrome.storage.sync`

ArnaudLigny:
Work in progress (issue #39).

To do:
-  [ ] handle default value by key (instead of global object): https://github.com/sindresorhus/github-notifier-chrome/blob/feature-39-chrome.storage.sync/extension/api.js#L46
- [ ] managing data storage migration: https://github.com/sindresorhus/github-notifier-chrome/blob/feature-39-chrome.storage.sync/extension/main.js#L110
- [ ] delete "console.debug" traces

### DIFF
--- a/extension/api.js
+++ b/extension/api.js
@@ -53,7 +53,7 @@
 					// temporary
 					data = defaults;
 
-					storage.get(data, function(items) {
+					storage.get(data, function (items) {
 						if (chrome.runtime.error) {
 							console.log(chrome.runtime.lastError);
 						} else {
@@ -65,7 +65,7 @@
 				set: function (data, callback, sync) {
 					sync = true;
 					var storage = sync ? chrome.storage.sync : chrome.storage.local;
-					storage.set(data, function() {
+					storage.set(data, function () {
 						if (chrome.runtime.error) {
 							console.log(chrome.runtime.lastError);
 						} else {
@@ -96,19 +96,17 @@
 		var participating;
 		var url;
 
-		window.GitHubNotify.settings.get(['oauthToken', 'useParticipatingCount', 'rootUrl'], function(items) {
+		window.GitHubNotify.settings.get(['oauthToken', 'useParticipatingCount', 'rootUrl'], function (items) {
 			url = items.rootUrl;
 			token = items.oauthToken;
-			participating = items.useParticipatingCount ? '?participating=true' : '';
-
 			if (!token) {
 				cb(new Error('missing token'));
 				return;
-			} else {
-				opts = {
-					Authorization: 'token ' + token
-				};
 			}
+			opts = {
+				Authorization: 'token ' + token
+			};
+			participating = items.useParticipatingCount ? '?participating=true' : '';
 
 			if (/(^(https:\/\/)?(api\.)?github\.com)/.test(url)) {
 				url = 'https://api.github.com/notifications';
--- a/extension/api.js
+++ b/extension/api.js
@@ -41,12 +41,11 @@
 
 		var api = {
 			settings: {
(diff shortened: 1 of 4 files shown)


extension/api.js
```diff
@@ -35,23 +35,48 @@
...
-					var item = localStorage.getItem(name);
-					if (item === null) {
-						return {}.hasOwnProperty.call(defaults, name) ? defaults[name] : undefined;
-					} else if (item === 'true' || item === 'false') {
-						return item === 'true';
-					}
-					return item;
+				get: function (data, callback, sync) {
+					sync = true;
+					var storage = sync ? chrome.storage.sync : chrome.storage.local;
```
sindresorhus: Why are you using both `sync` and `local`? Only `sync`, is good enough.

sindresorhus:
We should subscribe to the [`onInstalled`](https://developer.chrome.com/extensions/runtime#event-onInstalled) event and move the existing data from localStorage to the chrome sync when the user upgrades to the latest version.

extension/main.js
```diff
@@ -17,21 +17,23 @@
...
-
-			if (interval !== null && interval !== intervalSetting) {
-				window.GitHubNotify.settings.set('interval', interval);
-				period = Math.ceil(interval / 60);
-			}
-
-			// unconditionally schedule alarm
-			chrome.alarms.create({when: Date.now() + 2000 + (period * 60 * 1000)});
+			window.GitHubNotify.settings.get('interval', function (items) {
+				intervalSetting = parseInt(items.interval, 10);
```
sindresorhus: We should now save interval as a number, not a string.

extension/api.js
Narno: I agree. It was a test.

sindresorhus:
Let me know when this is ready for another review ;)

Narno:
Yep, I have to find some time to work on the to-do list...

smblott-github:
(None of my business, but...)

Would it not be simpler to leave the configuration options in `localStorage` and just use `chrome.storage.sync` for pushing changes between installations?

(Then there'd be no need to change the settings API.  You'd just have a little bit of new code which is isolated from everything else.)

Narno:
@smblott-github it's very clever, I love the idea! o/

sindresorhus:
> (None of my business, but...)

No worries. The beauty of open source is that anyone can comment their opinion.

> Would it not be simpler to leave the configuration options in localStorage and just use chrome.storage.sync for pushing changes between installations?

I don't really see how that's simpler. The migration script is a one time thing and makes the code cleaner in the long term. localStorage is also sub-optimal as it's a key string store without support for types, so we can't store numbers there.

smblott-github:
> I don't really see how that's simpler.

It's simpler because (if I understand the code correctly) you don't need to make any changes to the `settings` API, or to any of the code that uses that API.  So fewer changes, and fewer chances to mess up.

> localStorage is also sub-optimal as it's a key string store without support for types, so we can't store numbers there.

Just use `JSON.stringify` in `settings.set`, and `JSON.parse` in `settings.get`; then you can save any type of data.

sindresorhus:
I guess we have different ways on seeing it. I always prefer the clean code way even if it comes with an additional one time cost.

> So fewer changes, and fewer chances to mess up.

Cleaner code also does that.

sindresorhus:
@Narno ping :)

sindresorhus:
Closing for now. Happy to reopen again if you ever get to this @Narno ;)

Narno:
Yes, I don't have not enough time to contribute currently (new job). Thx.

Issue #50, opened 2015-12-17
Reuse existing notification tab


rojepp:
Great extension, thanks!

I'd like it if a click on the badge takes me to an existing notification tab and reloads it. I always have lots of notification tabs open after a while.

sindresorhus:
Good idea. PR welcome :)

corvannoorloos:
It took a little to figure what happened.
And to get used to, as the notifications tab was always last. (plus I often used it simply to open an additional GitHub page)

But, it has been good thus far. Thanks!

corvannoorloos:
Really paying attention to this, it doesn't always use the existing notifications tab or opens a new tab if not already available. Sometimes a different existing tab is being used (non-GitHub).

It has something to do with already opened GitHub tabs, but still unsure when this exactly happens.

sindresorhus:
@corvannoorloos Open an issue ;)

corvannoorloos:
@sindresorhus I was hoping to come up with a better explanation by now.

Issue #43, opened 2015-10-02
Click should always go to notification page...


smblott-github:
Please consider reverting (or fixing up) #32 (click when there are no notifications goes to GitHub home page).

Bad use case:

- I get an email notification that there's a new issue/comment.
- Hit `Ctrl-Shift-G` (which I bind to GitHub Notifier).
- But I'm taken to the GitHub page .... 'cause GHN hasn't yet figured out that there's a new notification.

This is a really bad UX.

Possible fixes:

- Wholly revert #32.
- Go to https://github.com/notifications?all=1 if there seems to be no new notifications.
- Add a dynamic check as to whether there are new notifications.

sindresorhus:
@Narno Thoughts?

sindresorhus:
And obligatory XKCD:

![every-change-breaks-someones-workflow](https://cloud.githubusercontent.com/assets/170270/10249995/9fdf2928-6952-11e5-8a3d-c4b265c530bb.png)

Narno:
Hello @smblott-github, thanks for your feedback.

OK, what is happening : GHN fetchs notifications asynchronously from the GH API while respecting the interval allowed between each call.
If you receive an e-mail about a new notification and clicks on the icon (or use a keybord shortcut) between 2 calls, the GHN counter is not yest up-to-date.

About your possible fixes proposals:
> Wholly revert #32.

I love this feature, I can't kill it! :-) (I think @sindresorhus has the last word)

> Go to https://github.com/notifications?all=1 if there seems to be no new notifications.

Not sure it would be usable for most of user, maybe as an option?

> Add a dynamic check as to whether there are new notifications.I like this idea, but I don't know what will be the impacts on performances and risk of API spam.

Maybe the best solution is working on the desktop notification feature? #25 :-)

michaelficarra:
I would revert #32. I do not like this behaviour.

sindresorhus:
Getting the notification count is fast. Why not just fetch the count when the user clicks the button before taking them to GitHub. That way we'll be sure to do the correct thing.

smblott-github:
> Getting the notification count is fast. Why not just fetch the count when the user clicks the button before taking them to GitHub.

That seems better than the status quo.

However, I'd still argue that GHN is not a bookmark for the github home page; it's for notifications.  So [this](https://github.com/notifications?all=1) would be a better alternative target.

[...]

PR #51, opened 2015-12-17
Reuse existing notifications tab if available

rojepp:
Fixes issues #50 and incidentally, #43 

Complete JS newb, so inspect thoroughly. :)

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -92,9 +92,9 @@
 
 		chrome.tabs.query({currentWindow: true, url: ghTab.url}, tabs => {
 			if (tabs.length > 0) {
-                ghTab.selected = true;
+				ghTab.selected = true;
 				chrome.tabs.update(tabs[0].id, ghTab);
-			} else if (typeof tab !== 'undefined' || tab.url === 'chrome://newtab/')
+			} else if (typeof tab !== 'undefined' || tab.url === 'chrome://newtab/') {
 				chrome.tabs.update(null, ghTab);
 			} else {
 				chrome.tabs.create(ghTab);
--- a/extension/main.js
+++ b/extension/main.js
@@ -85,18 +85,21 @@
 
 		const ghTab = {url};
 
-		if (window.GitHubNotify.settings.get('count') > 0) {
-			ghTab.url = `${url}notifications`;
-			if (window.GitHubNotify.settings.get('useParticipatingCount')) {
-				ghTab.url += '/participating';
-			}
+		
+		ghTab.url = `${url}notifications`;
+		if (window.GitHubNotify.settings.get('useParticipatingCount')) {
+			ghTab.url += '/participating';
 		}
 
-		if (typeof tab !== 'undefined' && (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === ghTab.url)) {
-			chrome.tabs.update(null, ghTab);
-		} else {
-			chrome.tabs.create(ghTab);
-		}
+		chrome.tabs.query({currentWindow: true, url: ghTab.url}, function(tabs) {
+			chrome.extension.getBackgroundPage().console.log(tabs);
+			if (tabs.length > 0) {
+ 				chrome.tabs.update(tabs[0].id, ghTab);
+				chrome.tabs.update(tabs[0].id, {selected: true});
+			} else {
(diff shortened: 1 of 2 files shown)


extension/main.js
```diff
@@ -85,18 +85,21 @@
...
-			if (window.GitHubNotify.settings.get('useParticipatingCount')) {
-				ghTab.url += '/participating';
-			}
+		
+		ghTab.url = `${url}notifications`;
+		if (window.GitHubNotify.settings.get('useParticipatingCount')) {
+			ghTab.url += '/participating';
 		}
 
-		if (typeof tab !== 'undefined' && (tab.url === '' || tab.url === 'chrome://newtab/' || tab.url === ghTab.url)) {
```
sindresorhus: This logic needs to be preserved.

extension/main.js
```diff
@@ -85,18 +85,21 @@
...
-			chrome.tabs.update(null, ghTab);
-		} else {
-			chrome.tabs.create(ghTab);
-		}
+		chrome.tabs.query({currentWindow: true, url: ghTab.url}, function(tabs) {
+			chrome.extension.getBackgroundPage().console.log(tabs);
```
sindresorhus: Leftover

extension/main.js
```diff
@@ -85,18 +85,21 @@
...
+			if (tabs.length > 0) {
+ 				chrome.tabs.update(tabs[0].id, ghTab);
+				chrome.tabs.update(tabs[0].id, {selected: true});
```
sindresorhus: Just use one call to .update

extension/manifest.json
```diff
@@ -10,7 +10,8 @@
 	},
 	"permissions": [
 		"alarms",
-		"activeTab"
+		"activeTab",
+		"tabs"
```
sindresorhus: Looks like this isn't needed: https://developer.chrome.com/extensions/tabs#manifest

extension/main.js
```diff
@@ -85,18 +85,21 @@
 
 		const ghTab = {url};
 
-		if (window.GitHubNotify.settings.get('count') > 0) {
```
sindresorhus: Why are you removing this?

rojepp: Because of issue #43.

extension/main.js
rojepp: Which logic? 'chrome://newtab/' gets replaced instead of a new tab? What does the 'undefined' check give us here? Thanks.

extension/main.js
rojepp: Removed.

extension/main.js
rojepp: Fixed.

extension/manifest.json
rojepp: We need `.url`, and it says:
> However, if you require access to the url, title, or favIconUrl properties of tabs.Tab, you must declare the "tabs" permission in the manifest

extension/main.js
rojepp: Fixed.

rojepp:
Thanks for the guidance. Like I said, JS newb. :)

extension/manifest.json
rojepp: Without 'tabs' in the manifest, it works when a notifications tab already exists, but when there is none, it seems to overwrite whatever tab I have open. To me that indicates that I should leave out the check for 'undefined'? I'll try that.

rojepp: Yes, If I don't check for undefined, it works without the tabs permission. Seems to me a bug in the chrome security model? How can I query all tabs for urls without the tabs permission? :/

extension/main.js
sindresorhus: Alright. But you still need to check the count in `if (window.GitHubNotify.settings.get('useParticipatingCount')) {`

See: https://github.com/sindresorhus/notifier-for-github-chrome/pull/49

sindresorhus:
Landed. Thank you :)

rojepp:
Hah, sorry. Forgot about this. Read the notification, didn't act on it immediately, then forgot. Glad it is sorted!

raysegantii:
Apparently with this change, clicking on the notifier would turn my left-most tab to GitHub notification page regardless whether that tab is originally on the notification page.

PR #52, opened 2015-12-18
Do not redirect to github home page.

### LINKED ISSUE #43: Click should always go to notification page...

smblott-github:
Clicking on the page icon (with this PR) always takes you to the same page, even if the notification count is `0`.

Fixes #43.

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -85,11 +85,9 @@
 
 		const ghTab = {url};
 
-		if (window.GitHubNotify.settings.get('count') > 0) {
-			ghTab.url = `${url}notifications`;
-			if (window.GitHubNotify.settings.get('useParticipatingCount')) {
-				ghTab.url += '/participating';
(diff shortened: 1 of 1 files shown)


sindresorhus:
Fixed in https://github.com/sindresorhus/notifier-for-github-chrome/pull/51.

PR #53, opened 2015-12-29
Travis CI: Test on Node.js v4.x.x and v5.x.x

SrinivasanTarget:


sindresorhus:
Same as https://github.com/sindresorhus/atom-linter-xo/pull/13#issuecomment-167800724 No point in changing it when it only targets `stable` as it does it for a reason.

Issue #60, opened 2016-03-04
Wrong tab choice


LeoColomb:
Hey!

I got a little issue since few days with Notifier for GitHub on Chrome: when I click on the extension button, my Feedly tab ![image](https://cloud.githubusercontent.com/assets/846943/13530352/3a673f20-e221-11e5-9f08-2b9933698409.png) is focused and moved to GitHub notification, instead of a new fresh tab.

![image](https://cloud.githubusercontent.com/assets/846943/13530039/9b3e1b7c-e21f-11e5-99df-451fc26830ae.png)

Tested:
* No other extension
* Chrome restarted
* Feedly tab refresh
* Feddly tab closed
    * Then the Trello tab ![image](https://cloud.githubusercontent.com/assets/846943/13530332/2483753e-e221-11e5-8337-b73f9b98a13c.png), in fact the second tab, suffer the bug
* Starting with Feedly tab focused
    * Then the Trello tab ![image](https://cloud.githubusercontent.com/assets/846943/13530332/2483753e-e221-11e5-8337-b73f9b98a13c.png), in fact the next tab, suffer the bug
* GitHub Notifications page opened in a new tab
    * Here everything is OK.

Do you have any idea of why? Thanks!

Chrome 49.0.2623.75 m (64-bit)

sindresorhus:
// @rojepp

LeoColomb:
Seriously, I don't understand: all pined tab are selected in the function. Chrome bug or what?

![image](https://cloud.githubusercontent.com/assets/846943/13534637/6eafa974-e236-11e5-8641-ad5ff6c247e5.png)

michaelficarra:
Has to be a chrome bug.

LeoColomb:
@sindresorhus @rojepp 
Are you sure full `tabs` permission is not require for this query?
https://github.com/sindresorhus/notifier-for-github-chrome/blob/222be9923f53121edcb3f123d9249130a4eada9c/extension/manifest.json#L13
https://developer.chrome.com/extensions/activeTab#what-activeTab-allows

rojepp:
@LeoColomb I would think so too from reading that, but if that were true, the behavior should be more deterministic? I still haven't seen this issue myself.. :/

LeoColomb:
I agree.

rojepp:
OTOH, read the first few paragraphs here: https://developer.chrome.com/extensions/tabs

>
[...]

PR #61, opened 2016-03-10
Add option to always open notifications page in a new tab

SEAPUNK:
Both a preference and a workaround to #60.


sindresorhus:
Thanks for the suggestion, but not interested in having an option for this. Adding an option is almost never the solution.

PR #66, opened 2016-03-13
Implement desktop notifications.

### LINKED ISSUE #12: Use Notifications API to get desktop notifications

### LINKED ISSUE #70: Switch from XHR to Fetch
In https://github.com/sindresorhus/notifier-for-github-chrome/blob/master/extension/api.js

Would make the code a lot cleaner.

https://developer.mozilla.org/en/docs/Web/API/Fetch_API

It's been in Chrome since v42.

YuriSolovyov:
This is another attempt to add desktop notifications based on work of @inikulin in #25.
The main feedback in original PR was about notification content and action we perform on notification click. 

With minimal changes, I was able to make it look like so:
image

Though I'd like to purpose a bit different style:
image

With all that there are still some questions left:
- what do we show if there are more than 1 notification?
  - list notification? 

  image
 - just say "there are N notifications for you" ? or "there is `foo` and N-1 other notifications for you"?

- we need to somehow track last notification timestamp, so we don't show same thing several times.
- on > 1 notification, where should click bring you?

Fixes #12 
Fixes #70 

/cc @sindresorhus


YuriSolovyov:
Idea: maybe if there > 1 notifications, we can add button like `Next` so user can cycle through them and `Mark all as read` to mark all at once?

sindresorhus:
> Though I'd like to purpose a bit different style:

Yeah. I like that style better.

> what do we show if there are more than 1 notification?

Shouldn't there be a Chrome notification for each GitHub notification?

YuriSolovyov:
> Shouldn't there be a Chrome notification for each GitHub notification?

I thought that could be quite annoying for active github users.

YuriSolovyov:
@sindresorhus ATM the only problem is to convert [GitHub API](https://developer.github.com/v3/activity/notifications/) url to the actual issue url.

YuriSolovyov:
For now, I found 2 ways of getting issue url of notification:

1 - We can get if from notification info, but it is hackish and not future proof:
```js
[apiRes.repository.html_url].concat(apiRes.subject.url.split('/').slice(-2)).join('/')
// -> "https://github.com/octocat/Hello-World/issues/123"
```

2 - We can make a request to GitHub API to get this info:
```
curl 'https://api.github.com/repos/octokit/octokit.rb/issues/123'
//     ^- this is same as apiRes.subject.url
```
gives:
```js
{
  "url": "https://api.github.com/repos/octokit/octokit.rb/issues/123",
  "repository_url": "https://api.github.com/repos/octokit/octokit.rb",
  "labels_url": "https://api.github.com/repos/octokit/octokit.rb/issues/123/labels{/name}",
  "comments_url": "https://api.github.com/repos/octokit/octokit.rb/issues/123/comments",
  "events_url": "https://api.github.com/repos/octokit/octokit.rb/issues/123/events",
  "html_url": "https://github.com/octokit/octokit.rb/pull/123",
  "id": 6130680,
  "number": 123,
  "title": "Implement repo assignees",
  // ...
}
```
and we need `html_url` from here.
I think it is ok to make this request when user clicks notification, so we send request lazily when needed.

YuriSolovyov:
@sindresorhus this is ready for review.

Narno:
Hello @YuriSolovyov, I'm testing your fork: feedback next week! :-)

YuriSolovyov:
Cool!
/cc @fstoerkle @gcochard

extension/main.js
```diff
@@ -1,6 +1,68 @@
...
+				window.GitHubNotify.settings.set('lastNotificationUpdate', notification.updated_at);
+				window.GitHubNotify.settings.set(notificationId, notification.subject.url);
+			});
+		}
+	}
+
+	function handleNotificationClicked(notificationId) {
+		const url = window.GitHubNotify.settings.get(notificationId);
+		if (url) {
+			fetch(url).then(res => {
```
sindresorhus:
```js
fetch(url).then(res => res.json())
```

extension/manifest.json
```diff
@@ -9,7 +9,8 @@
 		"128": "icon-128.png"
 	},
 	"permissions": [
-		"alarms"
+		"alarms",
+		"notifications"
```
sindresorhus: Shouldn't this be an optional permission?

YuriSolovyov:
@sindresorhus addressed feedback.
I noticed that even though [we are setting `optional_permissions`](https://github.com/sindresorhus/notifier-for-github-chrome/blob/master/extension/main.js#L120), this flag is never checked, should I remove it or make it work somehow?

Narno:
@YuriSolovyov it's a trick to avoid popin each time user refuse the new permission (true and false are !== undefined).

YuriSolovyov:
@Narno I guess I should add same thing for notifications?

Narno:
Yes, but if you find a class solution, that would be cool! ;-)

YuriSolovyov:
@Narno I don't quite get how/if user can grant this permission later on in this case...

Narno:
@YuriSolovyov yep, the only way to do that for the moment is to re-install the extension. I don't know if it's a real problem or not :-)

YuriSolovyov:
@Narno 
For now I just factored out this to Promise-based API.
We might want to add a setting like `"Allow querying my tabs"` or something

YuriSolovyov:
I'm also thinking about making use of [`reason` field](https://developer.github.com/v3/activity/notifications/#notification-reasons) - we can have a map of messages for every type of `reason` field, so we can add this info to notification like 
```
<message>
<repo>
New comment | You were mentioned | You were assigned | You are subsribed
```
... and so on
cc @Narno

YuriSolovyov:
Now I'm also fixing #70 here

YuriSolovyov:
Can anyone try 
```bash
curl -v -H "Authorization: token <token>" "https://api.github.com/notifications?per_page=1"
```
and tell me if it returns non-empty ETag header?

Matsv:
@YuriSolovyov curl output: http://pastebin.com/8a3iU4g2

YuriSolovyov:
@Matsv please replace `<token>` with real token, you can generate one in settings, make sure you added `notifications` permission

Matsv:
@YuriSolovyov Ok, one sec

Matsv:
@YuriSolovyov http://pastebin.com/dWJYnMAM with notications from the public-apis repo

YuriSolovyov:
Thanks, @Matsv, seems like I'm not the only one.
Ok, I sent a letter to Github, maybe I missed something.

YuriSolovyov:
linter is complaining about `snake_case`, but since it comes from rails app, it is kind of normal and we can't change it, so we have (at least) 2 options:
1. Somehow disable the rule atomically, just for 1 function
2. Replace obj with `Map`:
```js
// object:
const reasons = {
  subscribed:   'You are watching the repository',
  manual:       'You are subscribed to this thread',
  author:       'You created this thread',
  comment:      'New comment',
  mention:      'You were mentioned',
  team_mention: 'Your team was mentioned',
  state_change: 'Thread status changed',
  assign:       'You were assigned to the issue',
  default:      ''
};

// Map:
const reasons = new Map([
  ['subscribed',   'You are watching the repository'  ]
  ['manual',       'You are subscribed to this thread']
  ['author',       'You created this thread',         ]
  ['comment',      'New comment'                      ]
  ['mention',      'You were mentioned'               ]
  ['team_mention', 'Your team was mentioned'          ]
  ['state_change', 'Thread status changed'            ]
  ['assign',       'You were assigned to the issue'   ]
  ['default':      ''                                 ]
]);
```

Any ideas?

Also, I'd like to get some feedback about wording of these messages, maybe someone has better ideas, or it is fine for now?

gcochard:
You can always disable it with /* eslint-disable snake_case */ before and then /* eslint-enable snake_case */ at the end of your handling code.

YuriSolovyov:
This is ready for another round of review.

package.json
```diff
@@ -3,7 +3,7 @@
     "test": "xo && mocha"
   },
   "devDependencies": {
-    "chrome-stub": "^1.2.0",
+    "sinon-chrome": "^1.1.2",
```
gcochard: Just curious, why this change? It appears the usage is exactly the same in the tests. Are you planning on adding more to the tests that chrome-stub doesn't support?

extension/api.js
```diff
@@ -64,67 +35,105 @@
...
+	};
 
-		let url = window.GitHubNotify.settings.get('rootUrl');
+	window.GitHubNotify.getTabUrl = () => {
+		let rootUrl = window.GitHubNotify.settings.get('rootUrl');
 
-		if (!token) {
-			cb(new Error('missing token'), null, window.GitHubNotify.defaults.interval);
-			return;
+		if (/api.github.com\/$/.test(rootUrl)) {
```
gcochard: Should continue to use `(api\.)?github\.com` in the regex to support users who simply have `github.com` as their root URL.

extension/api.js
```diff
@@ -64,67 +35,105 @@
...
+			rootUrl = 'https://github.com/';
 		}
 
-		if (/(^(https:\/\/)?(api\.)?github\.com)/.test(url)) {
-			url = 'https://api.github.com/notifications';
-		} else {
-			url += 'api/v3/notifications';
+		const tabUrl = `${rootUrl}notifications`;
+		if (window.GitHubNotify.settings.get('useParticipatingCount')) {
+			return `${tabUrl}/participating`;
```
gcochard: Drop the slash here.

extension/api.js
gcochard: Nevermind. Scrolled down and see that it was moved from below. This should work as-is.

YuriSolovyov:
@gcochard you are right about stub changes, works ok with `chrome-stub`, I just remember there were some problems with it (can't remember what), anyway, I switched back to `chrome-stub`.

YuriSolovyov:
Btw, I don't think we need to stub XMLHttpRequest anymore...

YuriSolovyov:
Btw, there is a bit weird implementation detail about reason field in API response that can make things a bit confusing:
> Note that the reason is modified on a per-thread basis, and can change, if the reason on a later notification is different.

>For example, if you are the author of an issue, subsequent notifications on that issue will have a reason of author. If you're then mentioned on the same issue, the notifications you fetch thereafter will have a reason of mention. The reason remains as mention, regardless of whether you're ever mentioned again.

gcochard:
I think that also applies to the emails github sends out and users should expect that behavior. Once github fixes it, we get the fix for free.

If you do encounter any other issues with `chrome-stub`, let me know by filling an issue on it. I'm the author and I wrote it primarily for testing this extension so I have an interest in making it work for all of our use cases.

gcochard:
LGTM

YuriSolovyov:
I'd like to use it for a couple more days just to be sure

Narno:
Hello @YuriSolovyov, I'm (re)testing your branch since yesterday: notifications never appears, may be because the extension don't asks me for notifications permissions.

YuriSolovyov:
@Narno anything interesting in console?

YuriSolovyov:
@Narno can you check following things:
1. does API response contain `Last-Modified` header? if not there will be `null` in `lastModifed` key in local storage (something is very weird here: I had this header on work, but now it is absent on home PC)
2. notifications might not some just after upgrade, but at least on 2nd check they should come.
3. I'm not sure, but it seems like on linux notifications only appear when browser is focused, might be distro-dependent.

YuriSolovyov:
3. nah, appear just fine even without focus

sindresorhus:
ping @Narno :)

YuriSolovyov:
Status update: I was pretty busy over past couple weeks, but I'm going to back to it this weekend, there is indeed a bug here and I'm going to fix it.

YuriSolovyov:
@Narno I updated PR, should be better now, I'll use for couple more days and post final go/non-go later.

Narno:
Nice, I will try the new version this night!

Narno:
OK, now Chrome asks me the right thing when I checks the "Show desktop notifications" option. Good new!

Narno:
OK, after several tests, in different contexts, desktop notifications works really fine! \o/

YuriSolovyov:
@Narno can you recall if transition from version without notifications to this PR was ok?
Anything you did by hand that should've happen automatically?
Or switched just fine?

Narno:
Hello @YuriSolovyov, what I have tested:
* installing the PR extension in // of the current extension (with the same token): OK
* uninstalling the current extension then installing the PR extension (with the same token) : OK
* installing the PR extension on a fresh Chrome installation (with a new token) : OK

Done on my laptop computer (Mac OS X Yosemite) and on my desktop computer (Windows Seven).

YuriSolovyov:
@sindresorhus I think this is ready.

YuriSolovyov:
@sindresorhus now it's your turn :smile:

Narno:
:+1:

sindresorhus:
Awesome! Superb work @YuriSolovyov :)

Could you add something to the readme about the desktop notification feature?

I've published a new version. Should be live within an hour.

YuriSolovyov:
@sindresorhus sure, I'll add a couple words.

Issue #45, opened 2015-10-27
Icon shows maximum 50 unread notifications


rsalevsky:
The icon shows maximum 50 unread notifications even when they are more. 

![ghn](https://cloud.githubusercontent.com/assets/4128547/10766968/41eaf112-7cda-11e5-94c3-5d59a5ec9609.png)

I can reproduce this behavior every time, when I remember correct then this is the magic github pagination number.

techtonik:
Most likely a GitHub API limitation. What it could do is fetch and parse https://github.com/notifications HTML directly:
```
 <div class="columns">
        <div class="column one-fourth">
          <ul class="filter-list js-pjax-active" data-pjax>
            <li>
              <a href="/notifications"
                class="filter-item ">
                <span class="count">1583</span>
                Unread
              </a>
            </li>
            <li>
              <a href="/notifications/participating"
                class="filter-item selected">
                <span class="count">121</span>
                Participating
              </a>
            </li>
            <li>
              <a href="/notifications?all=1"
                class="filter-item ">
                All notifications
              </a>
            </li>
          </ul>
```

sindresorhus:
Yes, the GitHub API is paged. We would need to fetch the rest of the pages to show it all. I guess the easiest workaround would be to just show `50+` when it's 50.

techtonik:
That won't help me with 1583 notifications available. Is it possible to read query HTML page instead if the last state was 50+? Or store the value of last page?

sindresorhus:
It actually used to parse the HTML before the API was available. The problem with that is that you have to load all the HTML on each check and it's very fragile for GitHub changing around their HTML.

techtonik:
https://developer.github.com/guides/traversing-with-pagination/ lists that some API calls contain Link header, but that doesn't work for notifications. Have you tried to contact them to query that info?

Otherwise I think it would be ok to maintain code for HTML parsing.

sindresorhus:
> Have you tried to contact them to query that info?

No, I've given up on asking them for stuff. It never happens. I've wasted too much time writing them emails. Feel free though!

> Otherwise I think it would be ok to maintain code for HTML parsing.

PR welcome.

SamVerschueren:
Wouldn't it be better to write a more clever pager?

Just a quick example.

1. Page 1 is loaded and it appears I have more then 50 notifications
2. Page 10 is loaded but it does not exist
3. Page 5 is loaded and returns 50 notifications
[...]

PR #69, opened 2016-03-18
Show total notifications amount (fixes issue #45)

techtonik:
This uses new GitHub paging for notifications API, which passes total number of pages in Link header. It also use a trick to show 1 result per page to get number of notifications as a number of pages.

### DIFF
--- a/extension/api.js
+++ b/extension/api.js
@@ -30,10 +30,10 @@
 		};
 	})();
 
-	const getLinkPages = (linkheader) => {
-		for (let link of linkheader.split(', ')) {
+	const getLinkPages = linkheader => {
+		for (const link of linkheader.split(', ')) {
 			if (link.endsWith('rel="last"')) {
-				return Number(link.slice(link.lastIndexOf('page=')+5, link.lastIndexOf('>')));
+				return Number(link.slice(link.lastIndexOf('page=') + 5, link.lastIndexOf('>')));
 			}
 		}
 	};
@@ -115,10 +115,10 @@
 
 			if (data && data.hasOwnProperty('length')) {
 				const pages = getLinkPages(response.getResponseHeader('Link'));
-				if (pages == 1) {
+				if (pages === 1) {
 					cb(null, data.length, interval);
 				} else {
-					cb(null, (pages-1)*data.length, interval);
+					cb(null, (pages - 1) * data.length, interval);
 					// need one more query to fetch the number
 					// of notifications on the last page
 				}
--- a/extension/api.js
+++ b/extension/api.js
@@ -69,7 +69,7 @@
 		const opts = {
 			Authorization: `token ${token}`
 		};
-		const participating = window.GitHubNotify.settings.get('useParticipatingCount') ? '?participating=true' : '';
+		const participating = window.GitHubNotify.settings.get('useParticipatingCount') ? 'participating=true' : '';
 		let url = window.GitHubNotify.settings.get('rootUrl');
 
 		if (!token) {
@@ -83,12 +83,14 @@
 			url += 'api/v3/notifications';
 		}
 
+		url += '?';
 		url += participating;
+		url += '&per_page=1';
 
 		xhr('GET', url, opts, (data, status, response) => {
 			const interval = Number(response.getResponseHeader('X-Poll-Interval'));
 			const linkheader = response.getResponseHeader('Link');
-			
+
 			const getLinkPages = linkheader => {
 				for (const link of linkheader.split(', ')) {
 					if (link.endsWith('rel="last"')) {
@@ -115,15 +117,8 @@
 				return;
 			}
 
-			if (data && data.hasOwnProperty('length')) {
-				const pages = getLinkPages(response.getResponseHeader('Link'));
-				if (pages === 1) {
-					cb(null, data.length, interval);
-				} else {
-					cb(null, (pages - 1) * data.length, interval);
-					// need one more query to fetch the number
-					// of notifications on the last page
-				}
+			if (data) {
+				cb(null, pages, interval);
 				return;
 			}
 
--- a/extension/api.js
+++ b/extension/api.js
@@ -30,14 +30,6 @@
 		};
 	})();
 
-	const getLinkPages = linkheader => {
-		for (const link of linkheader.split(', ')) {
-			if (link.endsWith('rel="last"')) {
-				return Number(link.slice(link.lastIndexOf('page=') + 5, link.lastIndexOf('>')));
-			}
-		}
-	};
-
 	window.GitHubNotify = (() => {
 		const defaults = {
 			rootUrl: 'https://api.github.com/',
@@ -95,6 +87,16 @@
 
 		xhr('GET', url, opts, (data, status, response) => {
 			const interval = Number(response.getResponseHeader('X-Poll-Interval'));
+			const linkheader = response.getResponseHeader('Link');
+			
+			const getLinkPages = linkheader => {
+				for (const link of linkheader.split(', ')) {
+					if (link.endsWith('rel="last"')) {
+						return Number(link.slice(link.lastIndexOf('page=') + 5, link.lastIndexOf('>')));
+					}
+				}
+			};
+			const pages = getLinkPages(linkheader);
 
 			if (status >= 500) {
 				cb(new Error('server error'), null, interval);
--- a/extension/api.js
+++ b/extension/api.js
@@ -72,7 +72,7 @@
 		const query = [];
 		if (window.GitHubNotify.settings.get('useParticipatingCount')) {
 			query.push('participating=true');
-		};
+		}
 		query.push('per_page=1');
 
 		let url = window.GitHubNotify.settings.get('rootUrl');
@@ -88,7 +88,7 @@
 			url += 'api/v3/notifications';
 		}
 
-		url += '?' + query.join('&');
+		url += `?${query.join('&')}`;
 
 		xhr('GET', url, opts, (data, status, response) => {
 			const interval = Number(response.getResponseHeader('X-Poll-Interval'));
--- a/extension/api.js
+++ b/extension/api.js
@@ -30,6 +30,14 @@
 		};
 	})();
 
+	const getLinkPages = (linkheader) => {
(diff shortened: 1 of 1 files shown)


techtonik:
I am not proficient in JavaScript / ES6 so any feedback is appreciated. Maybe somebody can explain the hard stuff about all those braces and scopes?

extension/api.js
```diff
@@ -83,10 +83,22 @@
 			url += 'api/v3/notifications';
 		}
 
+		url += '?';
 		url += participating;
+		url += '&per_page=1';
```
gcochard: If `participating` is empty, the ampersand should not be there.

You can make the URL query args into an array and then join on `&` to make it easy to do:

```
query = []
if(participating){
  query.push(participating);
}
query.push('per_page=1');
url += query.join('&')
```

gcochard:
Very clever setting `per_page` to 1 to avoid having to do any extra fetches!

I'm not sure about the scope issue you were having with your function, both versions look like they should work to me. Some other maintainer might be able to explain it.

techtonik:
> Very clever setting per_page to 1 to avoid having to do any extra fetches!

All credit goes to @izuzak. =) My solution was to request API modification to allow `page=-1` for the last page.

techtonik:
> I'm not sure about the scope issue you were having with your function, both versions look like they should work to me. 

First I was trying to define function with `function` keyword at the same place it is now. And got linter error `Move function declaration to function body root  no-inner-declarations` https://travis-ci.org/sindresorhus/notifier-for-github-chrome/builds/116843932#L500

Then I tried to move it a level above and its name became invisible.

extension/api.js
```diff
@@ -83,10 +88,20 @@
...
+		url += `?${query.join('&')}`;
 
 		xhr('GET', url, opts, (data, status, response) => {
 			const interval = Number(response.getResponseHeader('X-Poll-Interval'));
+			const linkheader = response.getResponseHeader('Link');
+
+			const getLinkPages = linkheader => {
+				for (const link of linkheader.split(', ')) {
+					if (link.endsWith('rel="last"')) {
+						return Number(link.slice(link.lastIndexOf('page=') + 5, link.lastIndexOf('>')));
```
YuriSolovyov: Maybe can be rewritten as:
```js
const lastLink = linkheader.split(', ').find((link) => {
    return link.endsWith('rel="last"');
});
const count = Number(lastLink.slice(lastLink.lastIndexOf('page=') + 5, lastLink.lastIndexOf('>')));
```

extension/api.js
```diff
@@ -83,10 +88,20 @@
 			url += 'api/v3/notifications';
 		}
 
-		url += participating;
+		url += `?${query.join('&')}`;
 
 		xhr('GET', url, opts, (data, status, response) => {
```
YuriSolovyov: Does `data` paramerter here becomes paged too?
It would be unfortunate because I need it for desktop notifications

techtonik: It was always paged, GitHub just added info about total pages in Link header. Or I misunderstood the question?

extension/api.js
techtonik: `find` with lambda function looks cryptic, but I can compile this, so let me try.

extension/api.js
YuriSolovyov: The problem is that as long as we only need notifications count, this is perfectly fine, be if we need some of that data (#66), we're in a bad situation because we will need to perform additional requests.

techtonik: You need to perform additional requests only if last notification timestamp is changed. The plugin can set flag to fetch full first page on next API call if that desktop notification is needed. I have 2054 pending notifications and I am not a user for desktop notifications.

It also seems to me that polling API is not effective for desktop notifications stuff. I'd ask GitHub is they have some pubsub API for real-time notifications. For me it seems even more effective than hammering the server every minute.
https://developer.github.com/v3/activity/events/#list-events-that-a-user-has-received

YuriSolovyov: GitHub people say that no real-time stuff is available, so I have to be smarter about handling notification responses ... :disappointed:

YuriSolovyov: Here is the quote:
> I can't think of a smarter approach. If you need the total count of notifications -- you can get that with one request by setting the per_page to 1 and looking at the link header. But if you want to fetch all the notifications -- you'll need to go through all the pages. You can set per_page to 100 in that case to get as many notifications as possible in a single page. You could perhaps be smart about this and use the since parameter to fetch only notifications that were updated since the last time you fetched the notifications. In most cases, there will be less than 100 of those, so you'd be making only 1 request.

techtonik: If there is `since` parameter then I don't see any problems.

YuriSolovyov:
:+1:  LGTM

gcochard:
LGTM

YuriSolovyov:
@techtonik I assume you were using the extension with this PR for a while and it works ok, right?

techtonik:
@YuriSolovyov yes, running it in development mode for a few days already. Everything is smooth.

YuriSolovyov:
Thanks, @techtonik.
@sindresorhus are we going to release a new version with this PR?

sindresorhus:
Nah, we can wait until #66 has landed.

sindresorhus:
Thanks @techtonik :)

techtonik:
No problemo.

YuriSolovyov:
@techtonik, I'm testing this from master and I don't have `Link` header in my response.
Can you take a look?

YuriSolovyov:
```
λ curl -I -H "Authorization: token <token>" https://api.github.com/notifications?per_page=1
```
Gives me:
```http
HTTP/1.1 200 OK
Server: GitHub.com
Date: Sat, 26 Mar 2016 15:37:40 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 6125
Status: 200 OK
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4923
X-RateLimit-Reset: 1459007353
Cache-Control: private, max-age=60, s-maxage=60
Vary: Accept, Authorization, Cookie, X-GitHub-OTP
ETag: ""
Last-Modified: Sat, 26 Mar 2016 15:20:33 GMT
X-Poll-Interval: 60
X-OAuth-Scopes: notifications
X-Accepted-OAuth-Scopes: notifications, repo
X-GitHub-Media-Type: github.v3
Access-Control-Expose-Headers: ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval
Access-Control-Allow-Origin: *
Content-Security-Policy: default-src 'none'
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-XSS-Protection: 1; mode=block
Vary: Accept-Encoding
X-Served-By: 2c18a09f3ac5e4dd1e004af7c5a94769
X-GitHub-Request-Id: 2ED810D1:6B9D:D888A5E:56F6ACC4
```

techtonik:
Try it from browser - https://api.github.com/notifications?access_token=...&page=2

YuriSolovyov:
@techtonik same thing.

techtonik:
@YuriSolovyov that strange. Does extension from master work for you?

What is your browser?
Are there any proxies?
Have you tried regenerating the key?

This are my headers with replaced auth key.
```
HTTP/1.1 200 OK
Server: GitHub.com
Date: Sun, 27 Mar 2016 03:45:51 GMT
Content-Type: application/json; charset=utf-8
Transfer-Encoding: chunked
Status: 200 OK
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4992
X-RateLimit-Reset: 1459053564
Cache-Control: private, max-age=60, s-maxage=60
Vary: Accept, Authorization, Cookie, X-GitHub-OTP
ETag: W/""
Last-Modified: Fri, 18 Mar 2016 19:07:07 GMT
X-Poll-Interval: 60
X-OAuth-Scopes: notifications
X-Accepted-OAuth-Scopes: notifications, repo
X-GitHub-Media-Type: github.v3
Link: <https://api.github.com/notifications?access_token=...&page=3>; rel="next", <https://api.github.com/notifications?access_token=...&page=42>; rel="last", <https://api.github.com/notifications?access_token=...&page=1>; rel="first", <https://api.github.com/notifications?access_token=...&page=1>; rel="prev"
Access-Control-Expose-Headers: ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval
Access-Control-Allow-Origin: *
Content-Security-Policy: default-src 'none'
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-XSS-Protection: 1; mode=block
Vary: Accept-Encoding
X-Served-By: 8a5c38021a5cd7cef7b8f49a296fee40
Content-Encoding: gzip
X-GitHub-Request-Id: 5639F77F:2E05:FE39CFC:56F7576E
```

YuriSolovyov:
@techtonik Chrome 50.0.2661.49 beta-m (64-bit), Win 10, I also tried MS Edge and I still can't see Link header. I don't think it is browser-related, because plain curl has same issue and I also tried Win 7 PC and problem remained.

/cc @gcochard @fstoerkle can anyone confirm that extension from master works ok ?

techtonik:
@YuriSolovyov that's very strange. I would ask GitHub support.

YuriSolovyov:
@techtonik `Link` header is provided only if you have strictly > 1 pages. ATM this will cause an error:
```js
const linkheader = response.getResponseHeader('Link'); // -> null
const lastlink = linkheader.split(', ').find(link => { // can't get .split of null
    return link.endsWith('rel="last"');
});
```

gcochard:
Getting this error when I have 1 notification. Otherwise everything works when running from master.

```
api.js:96 Uncaught TypeError: Cannot read property 'split' of null
err @ api.js:96
xhr.onreadystatechange @ api.js:15
```

techtonik:
This should fix it.

PR #87, opened 2016-06-23
Still schedule alarm even after error.

### LINKED ISSUE #81: Notification count not updating
Hi all! In the last 2 weeks the extension sometimes doesn't update the notification count until I click on it. And once I click I think it refreshes for a while and stops again.
This has been happening to a colleague in the office too.
Is this happening to anyone else or is just us for some reason?

YurySolovyov:
Fixes #81

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -118,6 +118,8 @@
 		}
 
 		render(symbol, [166, 41, 41, 255], text);
+		// check again in a minute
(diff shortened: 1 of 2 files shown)


YurySolovyov:
I kept period of 1 min, but we might want to schedule it less often after errors, @sindresorhus what do you think?

gcochard:
LGTM

Issue #76, opened 2016-04-02
Improve test coverage


YurySolovyov:
This is kind of obvious, but we need more of those. Might be related to #75.
I wonder though if we need to somehow stub GitHub API calls, or we should just intercept that on extension level.

YuriSolovyov:
It would probably be cool to also migrate to Ava

sindresorhus:
Fixed by https://github.com/sindresorhus/notifier-for-github-chrome/issues/90.

PR #89, opened 2016-07-21
Fix query params in checkDesktopNotifications

marpo60:
* Fixes https://github.com/sindresorhus/notifier-for-github-chrome/issues/78

Looking at the [implementation](https://github.com/sindresorhus/notifier-for-github-chrome/blob/2.5.0/extension/api.js#L96-L103) of `buildQuery` it seems that the desired behaviour was to show only participating notifications if that option was enabled

This was not happening because `{ perPage:10,  participating: true }.toString() => perPage=10,participating=true`

Similar case for showing the notification count: https://github.com/sindresorhus/notifier-for-github-chrome/blob/2.5.0/extension/api.js#L107

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -45,7 +45,7 @@
 
 	function checkDesktopNotifications(lastModifed) {
 		const query = window.GitHubNotify.buildQuery({perPage: 100});
-		const url = `${window.GitHubNotify.getApiUrl()}?${query}`;
+		const url = `${window.GitHubNotify.getApiUrl()}?${query.join('&')}`;
 
 		window.GitHubNotify.request(url).then(res => res.json()).then(notifications => {
 			showDesktopNotifications(notifications, lastModifed);


YurySolovyov:
This is probably my silly mistake.
LGTM. Thank you!
/cc @gcochard @Narno

Narno:
Yep, looks good to me too, thanks @marpo60! 👍 

@YurySolovyov an another example of the pertinence of your issue #76! :-)

YurySolovyov:
Landed, thanks again.
@sindresorhus can you do a patch release so people get the fix?

@Narno working on some stuff that would hopefully make testing and support easier.

marpo60:
Great guys, thanks! 👍

sindresorhus:
I think we should switch to the new [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) instead of doing the query string stuff manually.

YurySolovyov:
@sindresorhus I can do this as part of #90

sindresorhus:
👍

sindresorhus:
New version published.

PR #90, opened 2016-07-22
Rewrite using modules and test with AVA

### LINKED ISSUE #75: Improve app structure
ATM it is just a bunch of functions that call each other.
I think we can do better by at least groupping them into related services (just plain objects, really).
This also might help with adding more tests, thought it is not mutually exclusive.

Any other ideas are welcome.

YurySolovyov:
2 files we had so far became too-damn-big, and nearly impossible to test, so I decided to split stuff into separate focused "modules-services" to make it more clear, readable and hopefully testable.
It seems to work, though I didn't done any deep testing.
Still need to do some stuff:
- [x] Bump Chrome version to 52 to support URLSearchParams and classes (and just in case).
- [x] Make sure chrome runtime event handlers are properly bound/called WRT `this`
- [x] Use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
- [x] Chrome 49+ enabled destructuring and rest params. Use where appropriate.
- [x] Extract more services? (Extracted: Badge, Tabs)
- [x] Write some actual tests. This should be much easier now, though I might need some advice.
  - [x] DefaultsService
  - [x] PersistenceService
  - [x] NetworkService
  - [x] PermissionsService
  - [x] BadgeService
  - [x] API
  - [x] TabsService
  - [x] NotificationsService
- [x] Test options page
- [x] Try from-the-scratch experience

Submitting it now to get some feedback.
/cc @Narno @sindresorhus @gcochard @fstoerkle 
Fixes #75 and #76


extension/src/api.js
```diff
@@ -0,0 +1,106 @@
...
+	class API {
+		constructor(persistence, networking, permissions, defaults) {
+			this.DefaultsService = defaults;
+			this.PersistenceService = persistence;
+			this.NetworkService = networking;
+			this.PermissionsService = permissions;
+		}
+
+		buildQuery(options) {
+			const params = new URLSearchParams();
```
gcochard: One concern here, do we want to require chrome 49 for such a small feature?

gcochard:
The `chrome-stub` module _should_ have everything necessary. If not, it's not difficult to add. LGTM so far, except for my previous bikeshedding comment.

YurySolovyov:
@Narno it would be nice if you could test from-the-scratch experience and confirm it works well.
@sindresorhus any AVA advices?

Narno:
Hello @YurySolovyov, thanks for you awesome job! :+1:

My first feedback after a fresh install of your branch (`YurySolovyov:ava`): I can't save my token (or any option).
```
options.js:38 Uncaught TypeError: this.writeValue is not a functiononChange
```

YurySolovyov:
@Narno thanks. Fixed. Not sure how to add test for it though.

Narno:
My user tests:
- [x] Installation
- [x] Configuration (token, options, etc.)
- [x] Permissions
- [x] Badge count update
- [x] Notifications
- [x] No Internet connection (badge icon)
- [x] Wrong token (badge icon)

YurySolovyov:
@Narno thanks a ton, really appreciate it.

package.json
```diff
@@ -1,24 +1,41 @@
 {
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/sindresorhus/notifier-for-github-chrome"
+  },
```
sindresorhus:
```
"repository": "sindresorhus/notifier-for-github-chrome",
```

package.json
```diff
@@ -1,24 +1,41 @@
...
-    "mocha": "*",
-    "mock-localstorage": "^0.1.3",
+    "moment": "^2.14.1",
     "sinon": "^1.15.4",
+    "url-search-params": "^0.5.0",
     "xo": "*"
   },
+  "ava": {
+    "verbose": true,
+    "files": "test/*-test.*"
```
sindresorhus: No need to manually specify this. It's in the default pattern.

package.json
```diff
@@ -1,24 +1,41 @@
...
     "istanbul": "^0.4.0",
-    "mocha": "*",
-    "mock-localstorage": "^0.1.3",
+    "moment": "^2.14.1",
     "sinon": "^1.15.4",
+    "url-search-params": "^0.5.0",
     "xo": "*"
   },
+  "ava": {
+    "verbose": true,
```
sindresorhus: I prefer the simple output.

sindresorhus:
I like the changes, but feels like an overuse of classes. There's no practical reason for `DefaultsService`, `NetworkService`, `PermissionsService`, and `TabsService` to be classes. Should just be modules exporting some simple functions. And don't pass them around. Just use them as globals. That will make it easier to eventually switch to ES2015 modules.

package.json
YurySolovyov: without it, it complains that `test/utils.js` does not contain tests:
```
✖ No tests found in test/utils.js, make sure to import "ava" at the top of your test file
```

YurySolovyov:
I agree that it might be an overuse of classes, but I wanted consistency.

> And don't pass them around. 

Passing them around makes dependencies clear, something we'll get for free with ES2015 modules, but we are not there yet.

> Just use them as globals.

I'd like to avoid using globals at all.

I totally agree that with ES2015 modules that will look much nicer, but we don't have them yet, so we have  a couple of options here:
- Add some kind of bundler to build the app, Rollup comes in mind, because it seems more lightweight than WebPack
- Land it as is and get back to it when ES2015 hit stable.

sindresorhus:
> Passing them around makes dependencies clear, something we'll get for free with ES2015 modules, but we are not there yet.

But it creates a lot of bloat and indirection. Just prefix with `window.` when you use globals from other files and it's IMHO clear. Modules will most likely come to Chrome later this year. So this is only temporarily. When they come we can just add some imports and done. With the current way, we'll have to refactor all the classes to not accept the dependencies.

> Add some kind of bundler to build the app, Rollup comes in mind, because it seems more lightweight than WebPack, and get rid of it when ES2015 modules hit stable.

Let's add Rollup then. At least better than passing dependencies into the constructors.

YurySolovyov:
TBH, I expected more from Rollup:
- Output is great.
- Maybe its just me, but I couldn't find a good way to separate main and options bundle.
I had to either use same bundle in both cases, or use 2 separate bundles with parts of the same code (used that).
- No way of extracting common code ATM.
- Minification is quite problematic too, cause current Uglify does not support ES6.

Webpack?

sindresorhus:
> Maybe its just me, but I couldn't find a good way to separate main and options bundle. I had to either use same bundle in both cases, or use 2 separate bundles with parts of the same code (used that).

We don't need to separate them like that. They are separate things and should require separate builds. They just each require what they need.

> No way of extracting common code ATM.

Not sure what that means?

> Minification is quite problematic too, cause current Uglify does not support ES6.

There's no point in minifying. This is an extension. Minification usually worsen the JS performance.

package.json
```diff
@@ -4,13 +4,17 @@
     "url": "sindresorhus/notifier-for-github-chrome"
   },
   "scripts": {
-    "test": "xo && ava"
+    "test": "xo && ava",
+    "build": "npm run build-main && npm run build-options",
+    "build-main": "rollup -i extension/main.js -o extension/out/main.js -f umd -n main",
```
sindresorhus: Use long flags instead of short unreadable ones like `-i`

extension/main.js
```diff
@@ -1,92 +1,93 @@
...
-	const notifications = new NotificationsService(persistence, api, defaults, tabs);
-
-	function handleInterval(interval) {
-		const intervalSetting = parseInt(persistence.get('interval'), 10) || 60;
-		const intervalValue = interval || 60;
-
-		if (intervalSetting !== intervalValue) {
-			persistence.set('interval', intervalValue);
-		}
+'use strict';
```
sindresorhus: No need for this. Rollup adds it in the final bundle.

YurySolovyov:
@sindresorhus with classes, I could easily sneak into services internals and stub them, but how do I do it with modules? they keep deps in closures, and I can't really fiddle with them, any advice?

YurySolovyov:
My guess is that I should stub underlying `window.*` APIs, but I might be wrong

YurySolovyov:
@sindresorhus 
I switched from rollup to webpack because I couldn't get AVA import src files properly.
I tried suggested workarounds from the repo, but I couldn't get it working.
Also fixed all the tests.

extension/src/badge-service.js
```diff
@@ -0,0 +1,38 @@
...
+	},
+
+	getErrorData(error) {
+		const title = DefaultsService.getErrorTitle(error);
+		const symbol = DefaultsService.getErrorSymbol(error);
+		return {symbol, title};
+	}
+};
+
+module.exports = BadgeService;
```
sindresorhus: I don't see why everthing has to be a `service`. Just name it `module.exports = badge`; and the file `badge.js`.

extension/src/badge-service.js
```diff
@@ -0,0 +1,38 @@
+const DefaultsService = require('./defaults-service.js');
+
+const BadgeService = {
+	render(text, color, title) {
+		window.chrome.browserAction.setBadgeText({text});
+		window.chrome.browserAction.setBadgeBackgroundColor({color});
+		window.chrome.browserAction.setTitle({title});
+	},
+
+	renderCount(count) {
```
sindresorhus: Just export these methods individually, like:

```js
exports.renderCount = count => {
  ...
};
```

Then there's no need to use `this`.

extension/src/defaults-service.js
```diff
@@ -0,0 +1,72 @@
...
+	getErrorTitle(error) {
+		return errorTitles.get(error.message) || errorTitles.get('default');
+	},
+
+	getErrorSymbol(error) {
+		return errorSymbols.get(error.message) || errorSymbols.get('default');
+	}
+};
+
+module.exports = DefaultsService;
```
sindresorhus: Same here. No need for `service`.

extension/src/defaults-service.js
```diff
@@ -0,0 +1,72 @@
...
+
+	getBadgeErrorColor() {
+		return colors.get('error');
+	},
+
+	getNotificationReasonText(reason) {
+		return notificationReasons.get(reason) || '';
+	},
+
+	getDefaultTitle() {
```
sindresorhus: Why are all these methods. They could just be exported as variables:

```js
exports.defaultTitle = defaults.get('title');
```

extension/src/defaults-service.js
```diff
@@ -0,0 +1,72 @@
...
+]);
+
+const DefaultsService = {
+	getDefaults() {
+		const obj = {};
+		defaults.forEach((value, key) => {
+			obj[key] = value;
+		});
+		return obj;
+	},
```
sindresorhus: Would be better to just expose the `defaults` Map here instead of converting it to an object.

extension/src/network-service.js
```diff
@@ -0,0 +1,25 @@
...
+			return Promise.reject(new Error('missing token'));
+		}
+
+		const headers = this.getHeaders(token);
+
+		return window.fetch(url, {headers});
+	}
+};
+
+module.exports = NetworkService;
```
sindresorhus: No need to name it `service`. `module.exports = request` would be better.

extension/src/network-service.js
```diff
@@ -0,0 +1,25 @@
+const PersistenceService = require('./persistence-service.js');
+
+const NetworkService = {
+	getHeaders(token) {
```
sindresorhus: No need to expose this. It can just be internal.

extension/src/network-service.js
```diff
@@ -0,0 +1,25 @@
...
+		/* eslint-disable quote-props */
+		return {
+			'Authorization': `token ${token}`,
+			'If-Modified-Since': ''
+		};
+		/* eslint-enable quote-props */
+	},
+
+	request(url) {
```
sindresorhus: No need for a class. Just expose this as `module.exports`.

extension/src/notifications-service.js
```diff
@@ -0,0 +1,67 @@
...
+			return API.makeApiRequest({url}).then(res => res.json()).then(json => {
+				const tabUrl = json.message === 'Not Found' ? API.getTabUrl() : json.html_url;
+				return TabsService.openTab(tabUrl);
+			}).then(() => {
+				return this.closeNotification(notificationId);
+			}).catch(() => {
+				return TabsService.openTab(API.getTabUrl());
+			}).then(() => {
+				return this.closeNotification(notificationId);
+			});
```
sindresorhus: The 3 `.then`'s above could use inline arrow functions.

extension/src/api.js
```diff
@@ -0,0 +1,83 @@
...
+
+		const lastlink = linkheader.split(', ').find(link => {
+			return link.endsWith('rel="last"');
+		});
+
+		const count = Number(lastlink.slice(lastlink.lastIndexOf('page=') + 5, lastlink.lastIndexOf('>')));
+		return Promise.resolve({count, interval, lastModifed});
+	},
+
+	makeApiRequest(options) {
```
sindresorhus: Get the `json` here, so every call to this method doesn't have to do it manually.

extension/src/notifications-service.js
```diff
@@ -0,0 +1,67 @@
...
+			title: notificationInfo.subject.title,
+			iconUrl: 'icon-notif-128.png',
+			type: 'basic',
+			message: notificationInfo.repository.full_name,
+			contextMessage: DefaultsService.getNotificationReasonText(notificationInfo.reason)
+		};
+	},
+
+	filterNotificationsByDate(notifications, lastModifed) {
+		const lastModifedTime = new Date(lastModifed).getTime();
```
sindresorhus: `lastModifed` => `lastModified`

extension/src/notifications-service.js
```diff
@@ -0,0 +1,67 @@
...
+		return notifications.filter(notification => {
+			return new Date(notification.updated_at).getTime() > lastModifedTime;
```
sindresorhus: Use inline arrow function.

extension/src/notifications-service.js
```diff
@@ -0,0 +1,67 @@
...
+		});
+	},
+
+	showNotifications(notifications, lastModifed) {
+		this.filterNotificationsByDate(notifications, lastModifed).forEach(notification => {
```
sindresorhus: Use a `for..of` loop.

package.json
```diff
@@ -1,24 +1,45 @@
 {
+  "repository": {
+    "type": "git",
+    "url": "sindresorhus/notifier-for-github-chrome"
+  },
```
sindresorhus:
```
"repository": "sindresorhus/notifier-for-github-chrome",
```

package.json
```diff
@@ -1,24 +1,45 @@
...
   "xo": {
     "esnext": true,
     "envs": [
-      "browser",
-      "mocha"
+      "browser"
+    ],
+    "ignores": [
+      "extension/out/*"
```
sindresorhus: `out` => `dist`

It's a more common name for it

package.json
```diff
@@ -1,24 +1,45 @@
...
-      "GitHubNotify"
+      "API",
+      "URLSearchParams",
+      "BadgeService",
+      "DefaultsService",
+      "PersistenceService",
+      "PermissionsService",
+      "NotificationsService",
+      "NetworkService",
+      "TabsService"
```
sindresorhus: I don't think we need these anymore. We now `require` them.

webpack.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+	entry: {
+		main: './extension/main.js',
```
sindresorhus: You can drop all the `./`

extension/manifest.json
```diff
@@ -4,7 +4,7 @@
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github-chrome",
 	"manifest_version": 2,
-	"minimum_chrome_version": "46",
+	"minimum_chrome_version": "49",
```
sindresorhus: Make it `52` just in case.

extension/main.js
```diff
@@ -1,210 +1,91 @@
...
-			window.GitHubNotify.requestPermission('tabs').then(granted => {
-				window.GitHubNotify.settings.set('tabs_permission', granted);
-				openTab(tabUrl, tab);
-			});
-		} else {
-			openTab(tabUrl, tab);
-		}
-	});
+chrome.runtime.onInstalled.addListener(handleInstalled);
+chrome.browserAction.onClicked.addListener(handleBrowserActionClick);
```
sindresorhus: You sometimes use `window.chrome` and sometimes `chrome.` I think we should use only `chrome`.

test/api-test.js
```diff
@@ -0,0 +1,183 @@
+import test from 'ava';
+import sinon from 'sinon';
+import utils from './utils';
+
+global.window = utils.setupWindow();
```
sindresorhus: This could just be `utils.setupWindow();`, and you could modify `global.window` in the function itself.

test/api-test.js
```diff
@@ -0,0 +1,183 @@
+import test from 'ava';
+import sinon from 'sinon';
+import utils from './utils';
```
sindresorhus: It's more common to use the singular form `util`.

test/tabs-service-test.js
```diff
@@ -0,0 +1,102 @@
...
+test('#openTab updates empty tab if provided', t => {
+	const service = t.context.service;
+	const url = 'https://api.github.com/resource';
+	const emptyTab = {id: 0, url: 'chrome://newtab/'};
+
+	window.chrome.permissions.contains = sinon.stub().yieldsAsync(true);
+	service.updateTab = sinon.spy();
+	service.queryTabs = sinon.stub().returns(Promise.resolve([]));
+
+	return service.openTab(url, emptyTab).then(() => {
```
sindresorhus: Don't return the Promise. Instead use `async t => {` on line 87 and `await `service.openTab`.

Applies to all the tests that does async.

You can then use the return value directly instead of in the `.then()` callback.

test/tabs-service-test.js
```diff
@@ -0,0 +1,102 @@
...
+import sinon from 'sinon';
+import utils from './utils';
+
+global.window = utils.setupWindow();
+
+const TabsService = require('../extension/src/tabs-service.js');
+
+test.beforeEach(t => {
+	t.context.service = Object.assign({}, TabsService);
+	global.window = Object.assign({}, utils.setupWindow());
```
sindresorhus: Why are you doing this here? It's already done on line 5.

test/permissions-service-test.js
```diff
@@ -0,0 +1,84 @@
...
+	});
+});
+
+test('#queryPermission returns rejected Promise if chrome.runtime.lastError is set', t => {
+	const service = t.context.service;
+
+	return utils.nextTickPromise().then(() => {
+		window.chrome.permissions.contains = sinon.stub().yieldsAsync();
+		window.chrome.runtime.lastError = new Error('#queryPermission failed');
+		return service.queryPermission('tabs').then(() => t.fail()).catch(() => t.pass());
```
sindresorhus: Should use `t.throws` (which supports promises).

sindresorhus: Applies to all places where you do this.

test/defaults-service-test.js
```diff
@@ -0,0 +1,84 @@
...
+	t.is(DefaultsService.getErrorSymbol({message: 'missing token'}), 'X');
+
+	const invalidMessages = [
+		'no such thing',
+		undefined,
+		NaN,
+		{foo: 312}
+	];
+
+	invalidMessages.forEach(message => {
```
sindresorhus: `for..of` loop

YurySolovyov:
@sindresorhus I think I've fixed most if not all the stuff.

YurySolovyov:
Just kind of generic note: classes were ok to start with just to create a picture of how I can break the monolith apart, but when it was done, there was no point in keeping them. Thanks for the tip, Sindre.

jacobbearden:
LGTM :+1:

sindresorhus:
@YurySolovyov I'm very sorry about such a slow reply. This looks good to me now :)

Can you fix the merge conflict?

YurySolovyov:
@sindresorhus sure, will do as I get home, thanks.

YurySolovyov:
@sindresorhus webpack can't build it on Linux Without this:
```diff
index 7de2610..41ef1eb 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,10 +1,10 @@
 module.exports = {
        entry: {
-               main: 'extension/main.js',
-               options: 'extension/options.js'
+               main: './extension/main.js',
+               options: './extension/options.js'
        },
        output: {
-               path: 'extension/dist',
+               path: './extension/dist',
                filename: '[name].js',
                publicPath: ''
        }
```

I think it ok to add it, right?

sindresorhus:
Yay! Glad we could finally land this. Amazing work @YurySolovyov :)

1ab7e29c-117e-11e6-9cbb-f0828bacea8e

Narno:
👍

gcochard:
:confetti_ball: :fireworks: :champagne:

YurySolovyov:
I doubt I can repeat anything like this anytime soon :)
But I'll do my best! Thanks everyone!

PR #103, opened 2016-10-22
Use type="password" for the OAuth token field

not-an-aardvark:
When using this extension, I've found it annoying that the OAuth token gets displayed whenever I open the options menu. It seems unnecessary to display the OAuth token that's currently being used, and it's also a mild security risk (the token is visible to anyone looking over my shoulder).

This PR changes the OAuth token field to use `<input type="password">`, to prevent the token from being displayed in plaintext.

### DIFF
--- a/extension/options.html
+++ b/extension/options.html
@@ -10,7 +10,7 @@
 			section h3 {
 				-webkit-margin-start: -18px;
 			}
-			input[type='text'] {
+			input[type='text'], input[type='password'] {
 				width: 350px;
 			}
 			span.description {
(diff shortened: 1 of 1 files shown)


not-an-aardvark:
The linter seems to be failing, but I don't think the failure is related to this PR.

techtonik:
And what if you want to know which token exactly do you use there?

not-an-aardvark:
What's the use-case for wanting to know that?

techtonik:
Clean up the token list.

not-an-aardvark:
But you can't actually see the value of the tokens in the [token list](https://github.com/settings/tokens) anyway, so being able to see a token in GitHub notifier isn't useful for cleaning up the list.

jacobbearden:
If you pull the latest changes and then push again the tests should pass.

I'm alright with merging this :+1:
If anyone else is we can go ahead with it.

not-an-aardvark:
Rebased onto master.

YurySolovyov:
I'm ok with it

techtonik:
> But you can't actually see the value of the tokens in the token list anyway, so being able to see a token in GitHub notifier isn't useful for cleaning up the list.

My bad. Confused it with SSH keys screen.

PR #112, opened 2016-12-24
WIP. Use storage.sync

### LINKED ISSUE #39: Use `chrome.storage.sync` instead of `localStorage`

YurySolovyov:
Fixes #39 
Also uses async/await and bumps chrome version to 55 to have native async/await support.
Babel is only needed to make it possible for webpack to get build passing because of async/await syntax, no transpilation should occur.
- [x] Fix tests
- [ ] Implement migrator
- [ ] Tests for migrator

Any comments are welcome.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -1,6 +1,6 @@
 {
 	"name": "Notifier for GitHub",
-	"version": "3.0.1",
+	"version": "3.0.2",
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github-chrome",
 	"manifest_version": 2,
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -4,12 +4,13 @@
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github-chrome",
 	"manifest_version": 2,
-	"minimum_chrome_version": "52",
+	"minimum_chrome_version": "55",
 	"icons": {
 		"128": "icon-128.png"
 	},
 	"permissions": [
-		"alarms"
+		"alarms",
+		"storage"
 	],
 	"optional_permissions": [
 		"tabs",
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -1,6 +1,6 @@
 {
 	"name": "Notifier for GitHub",
-	"version": "3.0.2",
+	"version": "3.0.1",
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github-chrome",
 	"manifest_version": 2,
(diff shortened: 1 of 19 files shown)


extension/manifest.json
```diff
@@ -1,15 +1,16 @@
 {
 	"name": "Notifier for GitHub",
-	"version": "3.0.1",
+	"version": "3.0.2",
```
sindresorhus: Don't bump this. Should be done only when actually publishing.

extension/src/persistence-service.js
```diff
@@ -1,30 +1,58 @@
...
 		}
 
-		return item;
+		return item[name];
 	},
 
 	set(name, value) {
-		window.localStorage.setItem(name, value);
+		return new Promise((resolve, reject) => {
+			chrome.storage.sync.set({[name]:value}, () => {
```
sindresorhus:
```js
chrome.storage.sync.set({[name]: value}, () => {
```

webpack.config.js
```diff
@@ -7,5 +7,16 @@ module.exports = {
...
 		filename: '[name].js',
 		publicPath: ''
+	},
+	module: {
+		loaders: [
+			{
+				test: /\.js$/,
+				loader: 'babel-loader',
+				query: {
+					"plugins": ["syntax-async-functions"]
```
sindresorhus: Single-quotes

sindresorhus (changes requested):

sindresorhus:
Nice! Any intention of migrating the old localstorage settings to `chrome.sync`? Would be annoying for users if they had to input their token again.

YurySolovyov:
I thought about that. For how long are you proposing to keep it?

sindresorhus:
> For how long are you proposing to keep it?

I assume you mean the migration code. It would be in a separate file, so we could just keep it for a long time. We would migrate the code once on update, so we can just completely switch to `chrome.storage`. Should use the [`onInstalled`](https://developer.chrome.com/extensions/runtime#event-onInstalled) even to detect an update and check if it's an older version that needs the migration and then run the migration script.

sindresorhus:
@YurySolovyov Is this something you think you'll finish eventually or should I close? I'm trying to clean up stale PRs/issues.

YurySolovyov:
Close this one, I'll reopen if/when I get back to it.

MasterOfTheTiger:
How far along is this?

YurySolovyov:
I think CI passed a couple of times, but I didn't implement migrator. This can probably be split into 3 parts:
- use async/await
- use storage.sync
- implement migrator from localstorage to storage.sync

I think it was my mistake in trying to do all at once.
If anyone finds anything of this useful, feel free to use it.

PR #123, opened 2017-04-12
Fix offline bug when connection restored & immediate update

TheBox193:
I found that very often when the internet connection was restored, the extension would stay in offline mode and not successfully wake back up. I believe this will fix that as well as have the added benefit of immediately getting the current notification count as time may have passed since we went offline.

I'm not fully certain why `scheduleAlaram()` doesn't effectively wake the extension. It should wake the extension 1 minute after the connection is restored via the alarm, but doesn't seem to. 

Nevertheless, it would be nice to update immediately when restored. This PR will do that via calling `update()` immediately after we come back online.

### DIFF
--- a/extension/main.js
+++ b/extension/main.js
@@ -81,7 +81,7 @@ function handleInstalled(details) {
 
 function handleConnectionStatus(event) {
 	if (event.type === 'online') {
-		scheduleAlaram();
+		update();
 	} else if (event.type === 'offline') {
 		handleOfflineStatus();
 	}


YurySolovyov:
I wonder what is wrong with tests...

TheBox193:
Yeah, I'll take a peek at the test later. Looks like quite a few failures. Good in concept, but that's concerning. 👍

TheBox193:
@YurySolovyov Re-ran tests on master, looks like it might be an issue there too. The previous run passed, but now we're failing. I don't think it relates to this PR. https://travis-ci.org/sindresorhus/notifier-for-github-chrome/builds/212652656

YurySolovyov:
Oh, well I'll take a look but can't promise it will be soon, feel free to investigate though.

TheBox193:
@YurySolovyov Rebased.

YurySolovyov:
LGTM :+1:

yimun:
lgtm

PR #124, opened 2017-04-19
Lock in static node version.

TheBox193:
Testing framework fails on later versions of node and should be updated.
This will lock node version at the latest stable point until updates to testing are done.


TheBox193:
I believe tests are failing do the the node version Travis is using.

It seems I'm unable to lock it in though.

gcochard:
Thanks fo this!

You'll want to edit the `travis.yml` file rather than `.nvmrc`. 

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Even-more-Nodejs-versions

gcochard:
I just looked at one commit and not the pr as a whole. You did the right thing but I believe Travis doesn't test changes to the Travis settings in PRs for security reasons. We will need to merge before it passes. I will verify and merge if confirmed.

.travis.yml
```diff
@@ -2,6 +2,7 @@ sudo: false
 language: node_js
 node_js:
   - 'node'
```
gcochard: Try removing this line as well

Issue #118, opened 2017-03-07
Installation on Firefox?


Cadair:
Hi, I am trying to swap out my extensions to enable multiprocess. I followed the link from the firefox repository to here but there is no obvious way to install this into firefox.

Can this extension be put into the Mozilla addon site?
Thanks!

sindresorhus:
> I followed the link from the firefox repository to here but there is no obvious way to install this into firefox.

Google is your friend.

> Can this extension be put into the Mozilla addon site?

If someone offers to do it or automates the process, sure.

Cadair:
> Google is your friend.

As always. I did manage to get it installed from the chrome webstore! My comment was really that your readme has no information on this for less technically savvy users. I will open a PR to the readme later.

Having got this installed in firefox, when I click on the link nothing happens I assume this is a bug? If so I will open a new issue.

ExE-Boss:
### I recommend looking at the following resources:
- https://developer.mozilla.org/Add-ons/WebExtensions/Porting_a_Google_Chrome_extension
- https://developer.mozilla.org/Add-ons/WebExtensions/Porting_a_legacy_Firefox_add-on
- https://developer.mozilla.org/Add-ons/WebExtensions/WebExtensions_and_the_Add-on_ID
- https://developer.mozilla.org/Add-ons/WebExtensions/Publishing_your_WebExtension

### Add-on ID and the minimum supported Firefox version
Since this addon has been published to AMO in the past, it is recommended that you set the Add-on ID to what it was in the legacy version.

This is achieved by adding the [`applications`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/applications) key to `manifest.json`:
```json5
"applications": {
	"gecko": {
		// This is the Add-on ID as it was
		// in the SDK version of the Add-on:
		"id": "jid1-DMjSwVBPnk7a1Q@jetpack",

		// This is the minimum Firefox version that supports
		// all the APIs needed by this WebExtension.
		// I have looked through the source code,
		// and it should be correct. (The `permissions` API
		// will be implemented in Firefox 55 (Bug 1197420))
		"strict_min_version": "55.0"

		// If you don’t care about the `permissions` API,
		// then you need to set `strict_min_version` to `52.0`
		// and create a dummy version of `PermissionsService`
		// for the Firefox version that auto-resolves
		// all Promises to `false` until Firefox 55.0
		// gets released.
	}
}
```

### Blocking bugs:
[...]

PR #128, opened 2017-05-24
Add `applications` key into `manifest.json` for proper Firefox support

ExE-Boss:
# ⚠️ Moved to #129

This extension uses APIs which are only available in Firefox 55+, and as such the `strict_min_version` has been set to `55.0`. See my comment on [#118](https://github.com/sindresorhus/notifier-for-github-chrome/issues/118#issuecomment-303750401) for details.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -33,7 +33,7 @@
 	"applications": {
 		"gecko": {
 			"id": "jid1-DMjSwVBPnk7a1Q@jetpack",
-			"strict_min_version": "55.0"
+			"strict_min_version": "52.0"
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


benrondeau:
@sindresorhus Are you accepting pull requests?

sindresorhus:
I'm not really interested in adding extra stuff for Firefox. The whole point of it supporting the Chrome extension API is that it should just work. I'm ok with https://github.com/sindresorhus/notifier-for-github-chrome/pull/128/commits/1f03dc577dc21fdb78bc4809f64fc94b58d81885 if you do a separate PR for it.

ExE-Boss:
@sindresorhus In that case, we have to wait for Firefox 55, which will add the [`permissions`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/permissions) API.

ExE-Boss:
PR moved to #129

PR #129, opened 2017-06-07
Make compatible with Firefox 55

ExE-Boss:
This extension uses APIs which are only available in Firefox 55+, and as such the [`strict_min_version`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/applications) has been set to `55.0`. Chrome does the exact same thing with [`minimum_chrome_version`](https://developer.chrome.com/extensions/manifest/minimum_chrome_version).

---

Moved from #128

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -29,5 +29,11 @@
 	"options_ui": {
 		"page": "options.html",
 		"chrome_style": true
+	},
+	"applications": {
+		"gecko": {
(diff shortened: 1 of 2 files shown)


EvertEt:
https://stackoverflow.com/a/37132274
https://github.com/sindresorhus/notifier-for-github-chrome/issues/132#issuecomment-307294996

PR #145, opened 2018-03-24
Require user to dismiss the interaction

cakenggt:
I often find that I don't know that I got new Github notifications when I was away from my desk, so making these require interaction makes them more like Slack's notifications.


kierenj:
Yes please!

gcochard:
I imagine quite a few people would be upset by this change. Can you make it a config option?

sindresorhus:
Yes, needs to be opt-in.

tooomm:
Is there a possibility to decide in a smart way?

Like if you are typing or moving the mouse then it will dismiss automatically after the normal time frame, because you are obviously sitting in front of the computer. If no input/movement is detected it'll stay.

sindresorhus:
Closing for lack of activity.

PR #151, opened 2018-04-26
Rewrite extension and fix all the things

### LINKED ISSUE #39: Use `chrome.storage.sync` instead of `localStorage`

### LINKED ISSUE #147: Simplify option handling
By using https://github.com/bfred-it/webext-options-sync

Example: https://github.com/sindresorhus/refined-github/blob/master/source/options.js

### LINKED ISSUE #148: Use `webextensions-polyfill`
So we can use promisified versions of the Chrome APIs

https://github.com/mozilla/webextension-polyfill

Example config to add it: https://github.com/sindresorhus/refined-github/blob/485488a3e886a29598948b75e399ef32860b7c0c/webpack.config.js#L26-L37

Example usage: https://github.com/sindresorhus/refined-github/blob/643a969bffda1c4a7eb4b8295603e3016e097186/source/features/mark-unread.js#L344

### LINKED ISSUE #149: Use `async/await`
Replace all the use of `.then()` with `await`.

Could also look into using other ES2015+ features.

notlmn:
Fixes #39 
Fixes #147 
Fixes #148 
Fixes #149

### Tasks:
- [x] Migrated to ESM
- [x] Use sync storage instead of `localStorage` (#39)
- [x] Handle form syncing with help of `webext-options-sync` (#147)
- [x] Using webextension-polyfill (#148)
- [x] Use `async`/`await` instead of promise callbacks (#149)
- [x] Removed services that are no longer needed (tests as well)
  - `network-request.js` (included in `api.js`)
  - `persistance-service.js` (using `browser.storage.local`)
- [x] Borrowed structure from refined-github
- [x] ~Using permission toggle for GHE (still doesn't work on Firefox)~
- [x] Simplify extension options page
- [x] Handle all errors (in options and in UI icon)
- [x] Test _everything_ manually
- [x] ~Add continuous deployment similar to RGH (for Firefox store)~ *https://github.com/sindresorhus/notifier-for-github/pull/151#issuecomment-385152467*
- [x] Make tests work

-------------------------

@sindresorhus review when possible, am I doing it all in one PR?


sindresorhus:
Awesome! I think you should focus on fixing the tests and getting this merged before doing additional features/changes. The PR is already pretty large.

sindresorhus:
> Add continuous deployment similar to RGH

This is already done: https://github.com/sindresorhus/notifier-for-github/blob/cfa3c7d2ac56c40666396f599585e0c55fd27fa4/.travis.yml#L7-L8

source/lib/local-store.js
```diff
@@ -0,0 +1,10 @@
+export default {
+	async get(key) {
+		const result = await browser.storage.local.get([key]);
+		return result[key];
+	},
+
+	async set(key, value) {
+		await browser.storage.local.set({key: value});
```
YurySolovyov: Did you mean `{ [key]: value }`?

notlmn: According to https://developer.chrome.com/extensions/storage, this is the format.

But I think the format you specified works too, for using the same value for multiple keys. 🤷

YurySolovyov: this is if you literally want `"key"` as the key. but since you're passing `key` as param, you probably want computed property

notlmn:
> > Add continuous deployment similar to RGH
>
> This is already done

But for Firefox store too!

notlmn:
I'm not very good at writing tests and haven't even tried to write before, now I had to learn a new testing framework (sinon).

I've tried all day long to fix the tests that are failing now, these are the tests that check if a function is called or not (only these types of tests are failing). And AVA just emits the body of the objects for each damn level, whenever a test fails.

I'm still trying to figure out why this is happening, but some pointers would help.

/cc @YurySolovyov.

YurySolovyov:
Are they timing out or fail in other ways?

sindresorhus:
> But for Firefox store too!

I don't intend to publish it on AMO. Their process is too nitpicky and annoying. Firefox users can easily get it through the Chrome store or the repo, per the readme instructions.

notlmn:
> Are they timing out or fail in other ways?

They are not timed out. All tests that test if a function is called or not fail (`stub#called`, `stub#calledWith`, `stub#callCount`).

image

YurySolovyov:
Are you taking into account that they (well, not tests themselves, but the APIs being tested) are likely async?
This means that you need to either `await` for the result before assertions or return a Promise.

notlmn:
> Are you taking into account that they (well, not tests themselves, but the APIs being tested) are likely async?

Yes, everything is running in async. Tests `await` for each API call. You can verify it by looking at the Travis logs.

Will look into it again!

YurySolovyov:
I have to admit that not all these tests are great and make a lot of sense.

Initially there were no tests at all, so I wanted to test everything when I wrote them.
But you don't have to keep them all just because of that - you can remove crappy ones and write better alternatives, just try not to remove them all :laughing:

notlmn:
@sindresorhus, @YurySolovyov the tests are modified and now the CI build passes. Removed a couple of tests that do not make sense as everything is async.

The tests look a bit scetchy thought as all of them rely on modifying `global.browser` object (some of the previous test did too).

Please review and tell me what you think.

test/notifications-service-test.js
```diff
@@ -1,171 +1,131 @@
...
-test('#openNotification closes notification if no url returned by PersistenceService', async t => {
-	const service = t.context.service;
-
-	window.localStorage.getItem = sinon.stub().returns(null);
-	service.closeNotification = sinon.stub().returns(Promise.resolve());
+test.serial('#openNotification skips network requests if no url returned by local-store', async t => {
+	const {service} = t.context;
 
-	await service.openNotification(t.context.notificationId);
+	await service.openNotification('random-notification-id');
```
YurySolovyov, line 69: That string is not random at all now, some of the test may pass by accident because of this.

notlmn: To pass the test, the argument here should be something that is not added to local store by `browser.storage.local.set()` before.

In the tests, the notifications are random numbers that are converted to strings. So  this would never collide with that.

This should've been something like `non-existent-notification-id`.

notlmn:
@sindresorhus @YurySolovyov can you review this again.

sindresorhus:
@notlmn Thanks for your patience. This looks really good now 👌

Would you be interested in joining the project as a maintainer? 🚀

notlmn:
Sorry, I didn't respond for too long.

The reason I haven't pinged you again is because someone very close to me died last month and I haven't got any time to be on GH for a month, and only got back last week.

I'd be very happy to join as a maintainer.

sindresorhus:
@notlmn Sorry to hear that, and no worries at all :)

sindresorhus:
@notlmn What should I put as your name in the readme?

notlmn:
> What should I put as your name in the readme?

"Laxman Damera" or just "Laxman" for short.

optimalgains719:
Andre Jones

Rowno:
Is there a reason why the "Read your browsing history" permission is now required? (It won't let you deny the permission)

image

It says here that all the permissions are optional:
image

Issue #155, opened 2018-07-25
Last update seems faulty


LebedevRI:
Hi. With google-chrome-beta, the last update (18.7.25.917) seems kinda faulty:
1. The icon is now generic bell, instead of github-like icon.
2. Clicking on it no longer does anything. It used to open github
3. Desktop notifications no longer appear to function.

I tried removing it and re-installing, to to avail.

mkurz:
+1
Same here. I miss the click behavior which opened https://github.com/notifications (in a new tab) before.
(I just wanted to open an issue but you were faster :wink:)

RobertoSchneiders:
I have the same problem on Chrome 67.0.3396.99 and 68.0.3440.75

ankurk91:
Same as #156

ghost:
https://github.com/sindresorhus/notifier-for-github/issues/155#issuecomment-407730964
It no longer opens the notifcations page in a new tab.
All the other things I can live with.

ravanscafi:
ping @sindresorhus

sindresorhus:
// @notlmn @YurySolovyov

YurySolovyov:
That moment when all tests are passing, yet your app is completely broken.

I guess the easiest way would be to revert recent changes and do a release.

notlmn:
Sorry for all the inconvenience the last update has caused. You can find the relevant info in the commit message.

TL;DR: I was too lazy to manually test after updating dependencies!

sindresorhus:
> The icon is now generic bell, instead of github-like icon.

That was a mistake. Fixed: https://github.com/sindresorhus/notifier-for-github/commit/67f9912a97aec1380d74497c776d2c2fb2869ed3

mkurz:
@notlmn @sindresorhus Thanks, however when I click on the icon now, I get a popup which wants me to confirm that the extension reads my browser history. Why?

mkurz:
Even after re-install the extension wants to get the "read browser history" permission...

LebedevRI:
Desktop notifications still seem iffy, too. Don't show up sometimes.

mkurz:
> Desktop notifications still seem iffy, too. Don't show up sometimes.

I can confirm that.

Why did you even update the extension? Was there a need? It was working great so far.

notlmn:
The notifications issue is fixed in the release `18.7.27.1445`.

mkurz:
[...]

PR #161, opened 2018-07-28
Remove redundant permissions from the manifest

notlmn:
Removes redundant or unnecessary permissions from the manifest.

Context: https://github.com/sindresorhus/notifier-for-github/issues/155#issuecomment-408549525, https://github.com/sindresorhus/notifier-for-github/issues/155#issuecomment-408590709

### DIFF
--- a/source/manifest.json
+++ b/source/manifest.json
@@ -10,14 +10,11 @@
 	},
 	"permissions": [
 		"alarms",
-		"storage",
-		"contextMenus"
+		"storage"
 	],
 	"optional_permissions": [
 		"tabs",
-		"notifications",
-		"http://*/*",
-		"https://*/*"
+		"notifications"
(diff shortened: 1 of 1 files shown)


vladholubiev (approved):

mkurz (approved):

YurySolovyov:
Do we need explicit permissions to be able to hit API?

notlmn:
> Do we need explicit permissions to be able to hit API?

These permissions are for accessing tabs, and modifying them (injecting CSS and JS). These are not required for making a network request (like touching an API).

YurySolovyov (approved):

YurySolovyov:
@notlmn any idea on how/if we can prevent requesting more permissions than needed in the future?

PR #162, opened 2018-07-30
Option: Always open notification page in new tab

mkurz:
If activated, the extension doesn't need the `tabs` permission (because there is no need to filter the tabs based on the url anymore).

See https://github.com/sindresorhus/notifier-for-github/issues/155#issuecomment-408634428 and following.

### DIFF
--- a/source/lib/tabs-service.js
+++ b/source/lib/tabs-service.js
@@ -1,4 +1,7 @@
+import OptionsSync from 'webext-options-sync';
 import {queryPermission} from './permissions-service';
+
+const syncStore = new OptionsSync();
 
 export const createTab = async url => {
 	if (browser.runtime.lastError) {
@@ -26,6 +29,10 @@ export const queryTabs = async url => {
 };
 
 export const openTab = async (url, tab) => {
+	const {newTabAlways} = await syncStore.getAll();
+	if (newTabAlways === true) {
+		return createTab(url);
+	}
 	if (await queryPermission('tabs')) {
 		const tabs = await queryTabs(url);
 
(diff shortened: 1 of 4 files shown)


source/lib/tabs-service.js
```diff
@@ -26,6 +29,10 @@ export const queryTabs = async url => {
 };
 
 export const openTab = async (url, tab) => {
+	const {newTabAlways} = await syncStore.getAll();
```
YurySolovyov, line 32: why not fetch `newTabAlways` specifically?

mkurz: Because in the rest of the codebase it's `getAll` everywhere. I just want to stay consistent.

YurySolovyov: okay, but it might be a good idea to go and fix such places someday, especially if such calls are inside event handlers or other UI-related code paths that can lead to lower perceived performance.

YurySolovyov:
@mkurz I assume you checked it locally / manually, and it works, right? Just to be sure.

mkurz:
@YurySolovyov Yes, sure I did.

YurySolovyov (approved):

mkurz:
Thanks for the fast review!

notlmn:
Thanks for the PR @mkurz 🎉.

mkurz:
You are welcome! Looking forward to the next release :+1:

notlmn:
> Looking forward to the next release

The extension is automatically pushed to store when we merge to master. Your changes are already live.

mkurz:
Thanks, I saw that. However I had to uninstall/install the extension again, because unfortunately there is no way to revoke permissions for an extension.
Thanks!

Issue #57, opened 2016-02-08
Use chrome.identity.launchWebAuthFlow to get an OAuth token


ImJasonH:
https://developer.chrome.com/apps/identity#method-launchWebAuthFlow

```
var CALLBACK_URL = 'https://'+chrome.runtime.id+'.chromiumapp.org';
var AUTH_URL = 'https://github.com/login/oauth/authorize/?client_id='+CLIENT_ID+'&redirect_uri='+CALLBACK_URL+'&scope=notifications';
...
// Opens a window to initiate GitHub OAuth, fires callback
// with token in the URL.
chrome.identity.launchWebAuthFlow({
  url: AUTH_URL,
  interactive: true,
}, function(redirectURL) {
  var q = redirectURL.substr(redirectURL.indexOf('#')+1);
  var parts = q.split('&');
  for (var i = 0; i < parts.length; i++) {
    var kv = parts[i].split('=');
    if (kv[0] == 'access_token') {
      token = kv[1];
      console.log('token is', token);
    }
  }
});
```

This would remove the need for users to generate a personal access token and copy/paste it in themselves.

sindresorhus:
:+1: Didn't know that was possible. Wanna do a pull request?

Use https://github.com/sindresorhus/query-string to get the query parameter. Browser version here: https://wzrd.in/standalone/query-string

YurySolovyov:
@ImJasonH what's `CLIENT_ID` in this case?

ImJasonH:
You can get a client ID by creating credentials for your project in the [Google Developers Console](https://console.developers.google.com/apis/credentials). Click "Create credentials" then select "OAuth client ID"

YurySolovyov:
@sindresorhus any thoughts? I might take a look at implementing this, but I think getting `CLIENT_ID` is available for extension owner only.

sindresorhus:
`CLIENT_ID` → `1053844927102-lb8knf98gq6o26bu4v5a3kpqpapj6tkl.apps.googleusercontent.com`

YurySolovyov:
@sindresorhus thanks, obviously blocked by #90 for now

sindresorhus:
#90 is now merged, finally ;)

YurySolovyov:
I'll see what I can do :)

ExE-Boss:
[...]

PR #166, opened 2018-10-28
[WIP] Use identity API flow to get an OAuth token

notlmn:
A dirty first look at #57, still a WIP.

@sindresorhus @YurySolovyov Do you think it's worth it?

### DIFF
--- a/source/options.js
+++ b/source/options.js
@@ -30,3 +30,60 @@ const update = async ({target: input}) => {
 for (const input of document.querySelectorAll('#options-form [name]')) {
 	input.addEventListener('change', update);
 }
+
+const button = document.querySelector('.js-auth-flow');
+button.addEventListener('click', async () => {
+	const options = await syncStore.getAll();
+	const url = new URL(options.rootUrl);
+	const origin = url.origin;
+
+	try {
+		const alreadyGranted = await queryPermission(origin);
+
+		if (!alreadyGranted) {
+			const granted = await requestPermission(origin);
+
+			if (granted) {
+				const CLIENT_ID = 'c209d287b799e3899f7a';
+				const CLIENT_SECRET = '2eab029032e28d0ca69b2cbf4332c813672f0003';
+				const CALLBACK_URL = browser.identity.getRedirectURL();
+				const SCOPES = ['notifications'];
+				const AUTH_URL = `https://github.com/login/oauth/authorize/?client_id=${CLIENT_ID}&redirect_uri=${encodeURIComponent(CALLBACK_URL)}&scope=${encodeURIComponent(SCOPES.join(','))}`;
+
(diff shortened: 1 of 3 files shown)


source/options.js
```diff
@@ -30,3 +30,60 @@ const update = async ({target: input}) => {
...
+
+	try {
+		const alreadyGranted = await queryPermission(origin);
+
+		if (!alreadyGranted) {
+			const granted = await requestPermission(origin);
+
+			if (granted) {
+				const CLIENT_ID = 'c209d287b799e3899f7a';
+				const CLIENT_SECRET = '2eab029032e28d0ca69b2cbf4332c813672f0003';
```
m1guelpf, line 48: Is there a danger in exposing the secret?

notlmn: As far as I can tell GitHub doesn't provide with implicit grant flow. You have to exchange a authorization token for access token.

And for this step you need client secret. These client secrets are usually stored on a proxy handled the access token step. But we do not have that architecture and all apps that have to store client secrets _somehow_ (according to the spec, the spec tells you nothing about storing client secrets).

GitHub however requires you to send the client secret while [auth flow](https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#2-users-are-redirected-back-to-your-site-by-github) to be sent in the URL parameters 😱🤯, which can be easily intercepted on a local network anyway, and I don't know what else to say about this.

If anyone knows a better way for storing client secrets on a public client, please do educate me (BTW, these are temporary tokens).

notlmn:
Closing this because to implement this, the extension would require permissions to all possible domains (`https://*/*`). We need to dynamically request permission to domains to which we want to fetch data from in order to get the ability to bypass CORS.

Considering that people do not want to give the extension more permissions, especially access to all the domains, and it is getting a lot more complex just to get a simple token.

I'll close the PR but leave the branch around in case anyone wants to poke around or find a solution.

source/options.js
ExE-Boss: > GitHub however requires you to send the client secret while auth flow to be sent in the URL parameters 😱🤯, which can be easily intercepted on a local network anyway

That is false, as HTTPS encrypts the path, query and fragment of the URL (the scheme is also technically encrypted, but the port number 443 with TLS encrypted payload tends to give away that it’s encrypted HTTPS traffic)

[The only part that is currently not encrypted in most cases is the hostname (the “api.github.com” part), but there’s work to encrypt that too.](https://en.wikipedia.org/wiki/Server_Name_Indication)

notlmn: > > GitHub however requires you to send the client secret while auth flow to be sent in the URL parameters 😱🤯, which can be easily intercepted on a local network anyway
>
> HTTPS encrypts the path, query and fragment of the URL

I'm pretty sure I was drunk when I wrote that (or late at night) 😅. I stand corrected!

PR #170, opened 2019-02-09
Check for all notifications page while opening new tab

### LINKED ISSUE #167: Add https://github.com/notifications?all=1 to open tab checker
Currently, the extension doesn't detect https://github.com/notifications?all=1 as open notifications page and always opens a new page on icon click instead of switching to already existing one.

notlmn:
Closes #167

### DIFF
--- a/source/lib/tabs-service.js
+++ b/source/lib/tabs-service.js
@@ -24,8 +24,13 @@ export const queryTabs = async url => {
 		throw new Error(browser.runtime.lastError);
 	}
 
+	const urlList = [url];
+	if (url.endsWith('/notifications')) {
+		urlList.push(url + '?all=1');
+	}
+
 	const currentWindow = true;
-	return browser.tabs.query({currentWindow, url});
+	return browser.tabs.query({currentWindow, url: urlList});
 };
 
 export const openTab = async (url, tab) => {


notlmn:
Already tested that it works, but pinging @romankornev for review anyway.

Will merge in 24 hours or before if no more issues arise.

source/lib/tabs-service.js
```diff
@@ -24,8 +24,13 @@ export const queryTabs = async url => {
...
 	}
 
+	const urlList = [url];
+	if (url.endsWith('/notifications')) {
+		urlList.push(url + '?all=1');
+	}
+
 	const currentWindow = true;
-	return browser.tabs.query({currentWindow, url});
+	return browser.tabs.query({currentWindow, url: urlList});
```
RomanKornev, line 33: That still opens the default URL.
It correctly finds `?all=1` tab and reloads it to `url` which is default here:
https://github.com/sindresorhus/notifier-for-github/blob/738ffbabd5d4bffd8ed22985adb6ec55a57303a5/source/lib/tabs-service.js#L45

RomanKornev (changes requested):

source/lib/tabs-service.js
notlmn: By definition, clicking on the extension icon should take you to the "notifications" page.

But when you have an "all notifications" page page or a "notifications" page open, this will take to the actual notifications page as intended by the click action on the extension icon.

Notifications page is where you generally go to find the recent/new notifications not the all notifications page.

RomanKornev: Can we add an option to customize that as well? Like "Use participating" that opens `/participating` but for `?all=1`. So that clicking from anywhere would open all notifications.

notlmn: Well we would end up having a lot of options for each little thing if we think of them that way.

IMO this should be fine, and not that much of a problem as most of the time people work in "current notifications" page.

I'll merge this for now, but if you really think it seriously is, then feel free to open an issue, but we can not guarantee to add more behavior because of request from one person.

🤞

PR #174, opened 2019-04-21
Jump to last unread comment

### LINKED ISSUE #169: Jump to newest comment instead of just opening the issue
That's how GitHub opens notifications.
`USER/REPO/issues/ISSUE_ID#issuecomment-COMMENT_ID`

notlmn:
Closes #169

After clicking a notification, based on its type we estimate the last unread comment URL based on
1. `last_unread_at`, may or may not exist, immediate comment of this is what we want
2. `updated_at`, used when `last_unread_at` is missing, gives us _last comment_
3. If both of the above fail, we just open the issue/PR as before

With this PR, we get handlers for each notification type. This allows us to handle notifications for types other than issues and pull-requests, which were ignored until now. (Need to implement other handlers, but issues and PRs should do fine for now)

### DIFF
--- a/source/background.js
+++ b/source/background.js
@@ -110,11 +110,6 @@ async function addNotificationHandler() {
 		browser.notifications.onClicked.addListener(id => {
 			openNotification(id);
 		});
-
-		// Possbily broken in browser, event listener getting fired if not displayed to user
-		// browser.notifications.onClosed.addListener(id => {
-		// 	removeNotification(id);
-		// });
 	}
 }
 
--- a/source/background.js
+++ b/source/background.js
@@ -4,7 +4,7 @@ import {openTab} from './lib/tabs-service';
 import {queryPermission} from './lib/permissions-service';
 import {getNotificationCount, getTabUrl} from './lib/api';
 import {renderCount, renderError, renderWarning} from './lib/badge';
-import {checkNotifications, openNotification, removeNotification} from './lib/notifications-service';
+import {checkNotifications, openNotification} from './lib/notifications-service';
 
 const syncStore = new OptionsSync();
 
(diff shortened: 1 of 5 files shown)


notlmn:
I've already tested this, but there's a change I missed something.

@MuntashirAkon as you are already using a modified version of this extension, can you test this PR, only if you can.

@RomanKornev have a look at this if you can, it would help a lot.

source/background.js
```diff
@@ -79,51 +81,59 @@ async function update() {
...
 async function addNotificationHandler() {
 	if (await queryPermission('notifications')) {
 		browser.notifications.onClicked.addListener(id => {
 			openNotification(id);
 		});
 
-		browser.notifications.onClosed.addListener(id => {
-			removeNotification(id);
-		});
+		// Possbily broken in browser, event listener getting fired if not displayed to user
```
YurySolovyov: either fix the issue, or remove the commented out code, we have complete file history in git for this.

source/lib/notifications-service.js
```diff
@@ -6,75 +6,119 @@ import localStore from './local-store';
...
 		}
+	} catch (error) {
+		throw error;
 	}
-
-	return false;
+}
+
+const notificationHandlers = {
+	/* eslint-disable quote-props */
```
YurySolovyov, line 55: if you're going to disable the rule, you might want to unquote the keys and disable whatever rule complains instead.

source/background.js
notlmn: I should've put this PR as a draft. This is intended as a test PR. It's late here so I thought I'd update it in the morning.

source/lib/notifications-service.js
notlmn: Disabling `quote-props` would only trigger `camel-case` next, so had to choose one of them. 🤷‍♂️

notlmn:
@YurySolovyov if you can do another review and confirm so that I can merge this quick.

There are some inconsistencies in the code base right now with the arrow functions, and the dependencies haven't been updated in some time. I'll send individual PRs for them later today.

(Also pinging @sindresorhus for any additional comments)

notlmn:
Merging this for now, feel free to open issues if you find any.

PR #175, opened 2019-04-24
Upgrade dependencies

notlmn:
- [x] Upgrade Node.js dependencies
- [x] Update tests accordingly
- [x] Switch from `uglifyjs` to `terser`


YurySolovyov (approved):
I'd drop semver modifiers from all packages. Otherwise, looks good to me.

PR #180, opened 2019-06-11
Setup publishing to AMO

### LINKED ISSUE #137: Stops working in Firefox 57
Hi, this plugin stopped working when I upgraded to FF 57 as Firefox removed the legacy api (ref https://support.mozilla.org/en-US/kb/firefox-add-technology-modernizing)

I'm willing to have a go at fixing this myself if it's not already in progress. Let me know if this is desirable :)

notlmn:
Setup Travis to publish extension to AMO. Follows the same publishing pattern as Refined GitHub except for the Git tags part.

The test extension resides at https://addons.mozilla.org/en-US/firefox/addon/nfg/.

Firefox support as of now is a bit finicky, like requesting permissions don't work yet, rest of the extension works fine though. Can be tested using the link above.

Fixes #137 (partially, for now), should make some Firefox users happy.

### Things to do
- [x] Handle Firefox' weird callback handling behavior
- [x] Fix tests (possibly rewrite them)
- [x] Setup daily publishing like RGH, not publish on push (can use hotfix tags for emergency releases, similar to RGH)

### Disclaimer

This is going to be a long PR, so Firefox users, hold your horses and if you can, provide feedback by testing the debug published to [AMO](https://addons.mozilla.org/en-US/developers/addon/nfg/versions).


source/background.js
```diff
@@ -123,7 +123,10 @@ function init() {
 	browser.runtime.onMessage.addListener(onMessage);
 	browser.runtime.onInstalled.addListener(handleInstalled);
 
-	browser.permissions.onAdded.addListener(addNotificationHandler);
+	// Chrome specific API
+	if (navigator.userAgent.includes('Chrome')) {
```
sindresorhus: This should be moved to a `util.js` file and be named `isChrome`, so we can use it elsewhere too. I think a better check would be `Boolean(window.chrome)`.

source/manifest.json
```diff
@@ -1,10 +1,16 @@
...
+	"version": "0.0.0",
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github",
 	"manifest_version": 2,
-	"minimum_chrome_version": "67",
+	"minimum_chrome_version": "69",
+	"applications": {
+		"gecko": {
+			"id": "{8d1582b2-ff2a-42e0-ba40-42f4ebfe921b}",
+			"strict_min_version": "62.0"
```
sindresorhus: I would prefer to target the latest version, which is 67

source/manifest.json
```diff
@@ -1,10 +1,16 @@
 {
 	"name": "Notifier for GitHub",
-	"version": "3.0.5",
+	"version": "0.0.0",
 	"description": "Displays your GitHub notifications unread count",
 	"homepage_url": "https://github.com/sindresorhus/notifier-for-github",
 	"manifest_version": 2,
-	"minimum_chrome_version": "67",
+	"minimum_chrome_version": "69",
```
sindresorhus: 74

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+	"presets": [
+		["@babel/preset-env", {
+			"targets": {
+				"chrome": "69",
+				"firefox": "62"
```
sindresorhus:
```suggestion
				"firefox": "67"
```

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+	"presets": [
+		["@babel/preset-env", {
+			"targets": {
+				"chrome": "69",
```
sindresorhus:
```suggestion
				"chrome": "74",
```

sindresorhus (changes requested):

ravanscafi:
I was able to install using provided URL :tada: 

However, I'm unable to mark the check "Show desktop notifications", even when opening the "new page".

image
(and also I think that the link to new page should have a `about="blank"`)

I'm willing to help, if it's wanted :)

ravanscafi:
I was able to install using provided URL :tada: 

However, I'm unable to mark the check "Show desktop notifications", even when opening the "new page".

image

(and also I think that the link to new page should have a `about="blank"`)

I'm willing to help, if it's wanted :)

sindresorhus:
I think we can actually just drop Babel. `esm` takes care of import/export.

ravanscafi:
Clicking on the button on bar has no effect :cry: 
(But it's correctly showing notification count badge tho).


Also, CSS is not loaded on options:
```
Security Error: Content at moz-extension://b8c0ffea-a9f9-4f79-87b7-0e0bd1fcbb2d/options.html may not load or link to chrome://mozapps/skin/extensions/extensions.css.
```

package.json
```diff
@@ -3,37 +3,45 @@
...
-		"lint-fix": "npm run lint:css -- --fix",
+		"lint:css": "stylelint source/**/*.css",
+		"lint-fix": "npm run lint:css -- --fix && npm run lint:js -- --fix",
 		"test": "npm run lint && npm run test:js && npm run build",
 		"test:js": "ava",
 		"build": "webpack --mode=production",
 		"watch": "webpack --mode=development --watch",
-		"release": "npm run build && npm run update-version && webstore upload --source=distribution --auto-publish",
-		"update-version": "VERSION=$(utc-version); echo $VERSION; dot-json distribution/manifest.json version $VERSION"
+		"prerelease:version": "VERSION=$(daily-version); echo $VERSION; dot-json distribution/manifest.json version $VERSION",
```
fregante: `daily-version` relies on git tags; you can use `utc-version`

source/background.js
fregante: If you wait a couple of weeks and comment on https://github.com/sindresorhus/refined-github/pull/2111, I’ll publish a polyfill for this soon. 😜

notlmn: > I think a better check would be `Boolean(window.chrome)`.

Unfortunately, `window.chrome` exists in Firefox too (for compatibility reasons), so  it had to be a UA sniff.

notlmn:
> However, I'm unable to mark the check "Show desktop notifications", even when opening the "new page".

> Clicking on the button on bar has no effect 😢
> (But it's correctly showing notification count badge tho).

These are all the same issues, for Firefox asynchronous calls from an event listener callback are not part of the same call-stack as the callback itself. So `browser.permissions.request()` fails saying that it should only called as part of user input, but as it is inside another async function, this fails.

Will have to rewrite a part of that logic, or someone can suggest me a workaround for this behavior in Firefox.

source/lib/permissions-service.js
```diff
@@ -1,19 +1,17 @@
...
 
-export async function requestPermission(permission) {
-	const granted = await browser.permissions.request({permissions: [permission]});
-
-	if (browser.runtime.lastError) {
-		throw new Error(browser.runtime.lastError);
+export function requestPermission(permission) {
+	try {
+		return browser.permissions.request({permissions: [permission]});
+	} catch (error) {
```
fregante, line 13: Why could this fail? I don’t think we need to catch it. `lastError` needed to be checked because without a check chrome will complain, but with promises this isn’t the case.

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+	"presets": [
+		["@babel/preset-env", {
+			"targets": {
+				"chrome": "74",
+				"firefox": "67"
+			}
+		}]
+	]
+}
```
fregante: As Sindre mentioned, Babel can probably be dropped altogether.

fregante (changes requested):
I think the solution is to *not* query the permissions. When you call `request` I think it will automatically pass, if we already have the permission, it won’t ask again. Untested but it wouldn’t make sense otherwise.

notlmn:
> I think we can actually just drop Babel. `esm` takes care of import/export.

Correct me if I'm wrong but Babel is a compile time dependency, whereas `esm` is a runtime dependency. Better to have the previous one than the later one as we are already using webpack to bundle stuff and little to no transpiling happening anyway.

notlmn:
> I think the solution is to _not_ query the permissions. When you call `request` I think it will automatically pass, if we already have the permission, it won’t ask again. Untested but it wouldn’t make sense otherwise.

The thing is that Firefox won't let you request permission in the first place, the function has to be called as part of a user interaction (which it already is), but according to Firefox, async function inside callback would not be considered as the same call-stack so Firefox thinks that the function is being called outside of user interaction. That is what I was explaining https://github.com/sindresorhus/notifier-for-github/pull/180#issuecomment-501556760.

source/lib/permissions-service.js
notlmn: It might fail because of https://github.com/sindresorhus/notifier-for-github/pull/180#issuecomment-501558819.

`runtime.lastError` is Chrome specific API, and I'm not seeing Chrome complaining about it, so dropped it.

.babelrc
notlmn: Can you comment on https://github.com/sindresorhus/notifier-for-github/pull/180#issuecomment-501557959. I'd like some feedback on that comment.

fregante:
We target the latest browsers, we can use the latest features without having to compile them down to ES5

notlmn:
Browserlist configuration in the `.babelrc` files makes sure that we are not compiling down to ES5. For example, the output file has async functions, they are not transpiled down to use generator functions.

I'll test dropping Babel altogether as we don't need transpiling anyway, and as we are bundling with webpack, I don't thing we need `esm` too.

fregante:
>  as we don't need transpiling anyway,

That’s what I’m saying 😅 
Babel was only there to add esm support but this PR introduced a whole preset-env

source/lib/permissions-service.js
fregante: This isn’t resolved, that part can be fixed. 

What I meant is that we in fact do not need to check that property because the Promised API polyfill does it automatically. 

The extension seems full of these checks in places where it doesn’t make sense. It could use a cleanup in a separate PR

source/lib/tabs-service.js
```diff
@@ -36,30 +36,30 @@ export async function queryTabs(urlList) {
...
-			}
-		}
+	if (newTabAlways) {
+		return createTab(url);
+	}
 
-		const matchingUrls = [url];
-		if (url.endsWith('/notifications')) {
-			matchingUrls.push(url + '?all=1');
+	const alreadyGranted = await queryPermission('tabs');
```
fregante: The mistake here is asking for permission “at runtime” instead of when the user checks the box. This check/request shouldn’t be here. 

The option itself seems backwards, it should be “Reuse notifications tabs when possible: default to false”

When the user checks this, the permission is requested. If the permission is denied we need to uncheck the box

fregante (changes requested):

source/lib/tabs-service.js
notlmn: > The mistake here is asking for permission “at runtime” instead of when the user checks the box. This check/request shouldn’t be here.

`openTab` almost always happens as part of a user input. According to the second example mentioned [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/User_actions), there cannot be any promises waited up before calling `browser.permissions.request()` (in Firefox only, Chrome is permissive).

I'm sure people at Firefox had a reason to do this, but I don't see how to solve this other than to revert #162, so we can drop `await` from [here](https://github.com/sindresorhus/notifier-for-github/blob/00d53edecf7587bd462fb8fc6b1716fc05b5eaeb/source/lib/tabs-service.js#L25).

------------------------

> The option itself seems backwards, it should be “Reuse notifications tabs when possible: default to false”

As mentioned, #162 can be reverted, @sindresorhus WDYT!?

------------------------

> When the user checks this, the permission is requested. If the permission is denied we need to uncheck the box

That already worked in Chrome, with https://github.com/sindresorhus/notifier-for-github/pull/180/commits/00d53edecf7587bd462fb8fc6b1716fc05b5eaeb, it now works in Firefox too.

fregante: The solution is not dropping it but following what I just said: don’t ask permissions when opening the tab

notlmn: The only place we request `tabs` permission is here 😕. Not asking permission here _or_ reverting #162, either of them require us to move `tabs` to required permissions (which is what I'm inclined towards, and need other people' opinion before making this change).

sindresorhus: `tabs` should stay optional. There are lots of people that would never want to allow it as it gives full access to all browsing history.

notlmn: Let's go with @bfred-it's suggestion then, to modify the existing option.

fregante: If you want to edits at a minimum, you can leave the current option as is and just move the `permissions.request` to the options page. 

The permission name is more of UX problem 🙂

notlmn:
Published `0.0.0.2` to AMO for final review. The extension should now work similar in both Chrome and Firefox.

Requesting a final review from @sindresorhus and @bfred-it.

ravanscafi:
Awesome! Notifications and clicking on the button on menu now are working!
Not sure about notification sound, though.

notlmn:
> Not sure about notification sound, though.

Yeah, was thinking about that too. I'm assuming it's probably something to do with autoplay blocking out of user interaction. I may be wrong, or someone can debug it for me.

ravanscafi:
I do believe that it was working on the very beginning, when we had issues with `browser.permissions.request()`.

notlmn:
> I do believe that it was working on the very beginning, when we had issues with `browser.permissions.request()`.

I can manually verify that both notifications and notification sounds work fine.

notlmn:
🚀 Successfully published to AMO as `19.6.16.428`, feel free to open new issues as they come.

PR #187, opened 2019-06-24
Fix PR URL returned by handler on API error

### LINKED ISSUE #185: opening browser notifications for pull requests
hey there is a problem when opening a pull request notification (browser notification) it redirect me to `https://github.com/org/repo/pulls/1757`

and that doesn't open the pull request for me 

it should be `https://github.com/org/repo/pull/1757` removing the `s` from `pulls`

TheSharpieOne:
Previously the URL for PRs would end up being similar to `https://github.com/user/repo/pulls/12345` which would direct the user to the PR list with the PR number as the author filter (which pretty much returns no results).
After: The URL will be `https://github.com/user/repo/pull/12345` which is the correct location.

Fixes #185


notlmn:
Thanks for the PR @TheSharpieOne 👍

PR #189, opened 2019-08-01
Update dependencies and deployment script

notlmn:
- Update dependencies, specifically `webext-options-sync`.
- Copy over build script to stay in sync with https://github.com/notlmn/browser-extension-template.
- Cleaned up some eslint and stylelint rules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
 language: node_js
 node_js:
-  - '12'
+  - stable
 env:
-  - EXTENSION_ID=lmjdlojahmbbcodnpecnjnmlddbkjhnn
+  - EXTENSION_ID='lmjdlojahmbbcodnpecnjnmlddbkjhnn'
 deploy:
-  provider: script
-  skip_cleanup: true
-  script: npm run release
-  on:
-    # Create a new release
-    #   On cron jobs, or
-    #   When triggered via "Trigger build" on https://travis-ci.org/sindresorhus/notifier-for-github
-    branch: master
-    condition: (($TRAVIS_EVENT_TYPE = api) || ($TRAVIS_EVENT_TYPE = cron && $(git rev-list -n 1 --since="26 hours ago" master)))
+  - provider: script
+    skip_cleanup: true
+    script: npm run release
+    on:
+      # On cron jobs https://docs.travis-ci.com/user/cron-jobs/
+      # When clicking "Trigger build" https://blog.travis-ci.com/2017-08-24-trigger-custom-build
+      branch: master
+      condition: (($TRAVIS_EVENT_TYPE = api) || ($TRAVIS_EVENT_TYPE = cron && $(git rev-list --since=yesterday HEAD)))
--- a/package.json
+++ b/package.json
@@ -1,32 +1,33 @@
 {
 	"private": true,
 	"scripts": {
-		"lint": "npm run lint:css && npm run lint:js",
+		"lint": "run-p lint:*",
 		"lint:js": "xo",
 		"lint:css": "stylelint source/**/*.css",
-		"lint-fix": "npm run lint:css -- --fix && npm run lint:js -- --fix",
-		"test": "npm run lint && npm run test:js && npm run build",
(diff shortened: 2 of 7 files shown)


source/options.js
```diff
@@ -5,10 +5,10 @@ optionsStorage.syncForm('#options-form');
 
 for (const inputElement of document.querySelectorAll('#options-form [name]')) {
 	inputElement.addEventListener('change', () => {
-		// `webext-options-sync` debounces syncing to 100ms, so send updates sometime after that
+		// `webext-options-sync` debounces syncing to 600ms, so send updates sometime after that
 		setTimeout(() => {
 			browser.runtime.sendMessage('update');
-		}, 200);
+		}, 700);
```
notlmn: @fregante, will be looking out for a new release of `webext-options-sync` to get rid of these arbitrary timeout values.

.travis.yml
```diff
@@ -1,15 +1,14 @@
 language: node_js
 node_js:
-  - '12'
+  - stable
 env:
-  - EXTENSION_ID=lmjdlojahmbbcodnpecnjnmlddbkjhnn
+  - EXTENSION_ID='lmjdlojahmbbcodnpecnjnmlddbkjhnn'
```
fregante, line 5: This isn't necessary, right?

package.json
```diff
@@ -1,45 +1,46 @@
...
-		"prerelease:source-url": "echo https://github.com/sindresorhus/notifier-for-github/tree/\"${TRAVIS_COMMIT:-$VERSION}\" > distribution/SOURCE_URL",
+		"prerelease:source-url": "if [[ $TRAVIS_REPO_SLUG ]]; then echo https://github.com/$TRAVIS_REPO_SLUG/tree/\"${TRAVIS_TAG:-$TRAVIS_COMMIT}\" > distribution/SOURCE_URL; fi",
 		"release": "npm-run-all build prerelease:* release:*",
 		"release:cws": "webstore upload --source=distribution --auto-publish",
 		"release:amo": "web-ext-submit --source-dir distribution"
 	},
 	"dependencies": {
 		"delay": "^4.3.0",
-		"webext-options-sync": "^0.16.0",
+		"webext-options-sync": "^0.21.2",
```
fregante: You can try the latest, while you're at it

```suggestion
		"webext-options-sync": "^1.0.0-4",
```

```sh
npm i webext-options-sync@next
```

source/options.js
fregante: Indeed, you can use the version I suggested and then:

```js
select('form').addEventListener('options-sync:form-synced', () => {
	browser.runtime.sendMessage('update');
});

File: package.json [hunk 4]
```diff
@@ -1,45 +1,46 @@
...
 		"webextension-polyfill": "^0.4.0"
 	},
 	"devDependencies": {
-		"ava": "^2.1.0",
+		"@types/chrome": "0.0.86",
```
fregante, line 24:
```suggestion
```

Should be dropped, there's no typescript here

source/options.js
notlmn: Yup, was trying to use that and soon found that this wasn't published to stable yet.

fregante: > Yup, was trying to use that and soon found that this wasn't published to stable yet.

It will be soon enough, it'd be good to test it here as well.

package.json
notlmn: Trying to stay on stable for now, planning on a rewrite once everything is available.

Love the new additions on `webext-options-sync` though. 😄

package.json
notlmn: So that I we can get code suggestions in VSCode. 😋

.travis.yml
notlmn: Unless there are any spaces, it's not. A matter of personal preference, if I can say.

Not seriously attached to it though, open to drop it if it is that big of an issue. 🤷‍♂️

source/options.js
notlmn: https://github.com/fregante/webext-options-sync/issues/36

PR #192, opened 2019-10-05
feat(filter): Filter notifications based on owner/repository

### LINKED ISSUE #190: feature: allow filtering for organization
It would be cool if users could set a list of organizations to receive the notifications from. When this list is set the extension should show the count only for those organizations specified.

#### Reasoning

I am using multiple browser profiles for my work and personal use and in my work profile, I am only concerned about the notification in specific organizations related to my work.

jroehl:
# This is a work in progress and meant as a first POC

pr

This PR resolves #190.

Added a new option in the extension options to filter notifications based on ownership or repository name. The repositories are fetched and stored in `chrome.storage.sync` for cross-browser use. 

Currently, there is an issue if the user is subscribed to a large number of repositories.
This [PR](https://github.com/fregante/webext-options-sync/pull/40) in [webext-options-sync](https://github.com/fregante/webext-options-sync/) will help by compressing the data before storing it.  

## TODO

- [ ] Get [PR](https://github.com/fregante/webext-options-sync/pull/40) merged
- [ ] Bump [webext-options-sync](https://github.com/fregante/webext-options-sync/) version
- [x] Test in FireFox
- [x] Write tests


source/repositories.js
```diff
@@ -0,0 +1,138 @@
...
+
+	document.querySelector('.repo-wrapper').innerHTML = html;
+}
+
+function getListMarkup(owner, repositories) {
+	const repos = Object.keys(repositories);
+
+	const list = repos
+		.map(repository => {
+			return `
```
notlmn, line 72: Instead of using custom HTML and JS code to toggle element visibility, we could use the native HTML [Details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) and [Summary](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) elements to achieve the same without any additional JS or CSS.

This was the same I used in my [original comment](https://github.com/sindresorhus/notifier-for-github/issues/190#issuecomment-520210530) for mockup. I think with that change, we can remove a significant amount of code in this file.

source/util.js
```diff
@@ -23,3 +28,21 @@ export async function isNotificationTargetPage(url) {
 	// Issue, PR, commit paths, and per-repo notifications
 	return /^(((issues|pull)\/\d+(\/(commits|files))?)|(commit\/.*)|(notifications$))/.test(repoPath);
 }
+
+export function parseLinkHeader(header) {
+	return header.split(',').reduce((links, part) => {
```
notlmn: This keeps failing, as `header` is `null` when user has less than 100 repositories.

Something like `parseLinkHeader(header = '')` should handle it.

source/util.js
```diff
@@ -23,3 +28,21 @@ export async function isNotificationTargetPage(url) {
...
+		const [sectionUrl, sectionName] = part.split(';');
```
notlmn:
```suggestion
		const [sectionUrl = '', sectionName = ''] = part.split(';');
```

Same reason as [above](https://github.com/sindresorhus/notifier-for-github/pull/192#discussion_r331778082).

source/repositories.js
```diff
@@ -0,0 +1,138 @@
...
+import {listRepositories} from './lib/repositories-service';
+import {getUser} from './lib/user-service';
+
+const form = document.querySelector('#repositories-form');
+const button = document.querySelector('#reload-repositories');
+const filterCheckbox = document.querySelector('[name="filterNotifications"]');
+
+button.addEventListener('click', () => !button.classList.contains('loading') && init(true));
+
+filterCheckbox.addEventListener('click', async () => {
```
notlmn: Any reason for not listening on the `change` event here? 🤔

source/lib/repositories-service.js
```diff
@@ -0,0 +1,47 @@
...
+import {makeApiRequest} from './api';
+
+export async function getRepositories(
+	repos = [],
+	params = {
+		page: '1',
+		per_page: '100' // eslint-disable-line camelcase
+	}
+) {
+	const {headers, json} = await makeApiRequest('/user/repos', params);
```
notlmn: This should be wrapped inside a `try {} catch {}` block, as we cannot know when and why a request might fail.

Also bubbling up the error to the user would be a massive UX as well. 👍

notlmn (changes requested):
I really love this feature, but here are some changes. 😉

notlmn:
Also can we try storing repository ids in local storage instead of repository names themselves, which might decrease the amount of storage space utilized by the extension. 🤔

jroehl:
> Also can we try storing repository ids in local storage instead of repository names themselves, which might decrease the amount of storage space utilized by the extension. 🤔

Yeah, thought about that - but then - every time the options are opened, the names would need to be fetched and mapped. Don't know if that's a bit too slow. 🤷‍♂

We could fetch it once a while and store it in local storage though - but that would mean the data is stored in multiple locations. Don't know what's best.

source/lib/repositories-service.js
jroehl: Very true - but what exactly do you mean by `would be a massive UX as well`?

source/util.js
jroehl: True - but as the header is really `null` sometimes I am using `(header || '')` instead of a default param value.

source/repositories.js
jroehl: That is very true - I wasn't fully aware of those semantic tags 🙈 
It does at least remove quite a lot of CSS. 🎉

source/lib/repositories-service.js
notlmn: If for any reason, the requests fail or cache fails to load, we could inform the user to something like "try again".

notlmn:
> Yeah, thought about that - but then - every time the options are opened, the names would need to be fetched and mapped. Don't know if that's a bit too slow. 🤷‍♂
> 
> We could fetch it once a while and store it in local storage though - but that would mean the data is stored in multiple locations. Don't know what's best.

I see your point, let's proceed with repo names then.

source/util.js
```diff
@@ -23,3 +28,21 @@ export async function isNotificationTargetPage(url) {
...
+		const [sectionUrl = '', sectionName = ''] = part.split(';');
+		const url = sectionUrl.replace(/<(.+)>/, '$1').trim();
+		const name = sectionName.replace(/rel="(.+)"/, '$1').trim();
+		return Object.assign({}, links, {[name]: url});
+	}, {});
+}
+
+export const background = {
+	/* eslint-disable no-undef */
+	log: chrome.extension.getBackgroundPage().console.log,
```
notlmn: These should be `browser.extension` right?

jroehl:
Firefox is tested and tests for some of the functions are in place. The only thing that's missing is the compression [PR](https://github.com/fregante/webext-options-sync/pull/40). Maybe you could try to bump this one as well 🙏 .

notlmn:
Looks good, but will review once the compression PR on that repo is merged, and when I get time to test this out myself. Good job.

---------------

> Maybe you could try to bump this one as well 🙏 .

Federico seems to be a bit inactive on that repo, will have to wait for him to be back.

fregante:
Perhaps this is _too advanced._ Wouldn't it be easier to only store muted repos?

As for the UI:


- display the list of muted repos
- display the last X repos that received a notification, so the user can mute them (this probably doesn't need to be stored at all. Just keep it in memory in background.js)

jroehl:
@fregante
> Perhaps this is _too advanced._ Wouldn't it be easier to only store muted repos?
> 
> As for the UI:
> 
> * display the list of muted repos

But if we store only the muted repos, we'd have to refetch the repos every time we open the options.
It's similar to [this](https://github.com/sindresorhus/notifier-for-github/pull/192#issuecomment-541748998) answer.

> * display the last X repos that received a notification, so the user can mute them (this probably doesn't need to be stored at all. Just keep it in memory in background.js)

If I understood correctly, this would mean the user would need to enable/disable repos every time when the browser is restarted. 🤔

fregante:
Not at all. To mute certain repos, you only need to know their names. We don’t need to display the rest of them, so you don’t need to fetch them either. 

An “enabled” repo is a repo that hasn’t been muted. Muted repos are stored, the rest aren’t. “Last 10 repos that received a notification” are only useful for the user to go through and mute them, so it’s only one fetch when they open the browser for the first time (and then they’ll stay in memory)

The key difference is that instead of having to fetch and store possibly hundreds of repos, you just store a few muted repos (and muted orgs)

jroehl:
Gotcha. IMHO the solution to show all repos and have fine granular control over them would be more interesting to me - as I don't want to mute the repos after I received a notification, but disable them beforehand. 

If I am not mistaken I'd still need to fetch all repositories and see if there are new notifications - to list the last X of them. So it'd still be some fetching to do. Or is there a repo endpoint that can be filtered by notifications?

But if this is the solution you want for this extension I could look into a way of doing this...

fregante:
> If I am not mistaken I'd still need to fetch all repositories and see if there are new notifications - to list the last X of them

Instead of fetching the repos, you just fetch the notifications, which we already do somewhere.

> But if this is the solution you want for this extension I could look into a way of doing this...

I'm not a developer here, I just try to suggest _lighter_ solutions that won't end up hitting storage limits and are easier to maintain. The `sync` storage is limited to [8KB per item](https://developer.chrome.com/extensions/storage#property-sync).

fregante:
I just realized: you can receive notifications from _any_ repo, so this list isn't complete to begin with. 😅

source/lib/notifications-service.js
```diff
@@ -122,8 +124,18 @@ export async function playNotificationSound() {
...
-	const {showDesktopNotif, playNotifSound} = await optionsStorage.getAll();
+	let notifications = await getNotifications({lastModified});
+	const {showDesktopNotif, playNotifSound, filterNotifications} = await optionsStorage.getAll();
+
+	if (filterNotifications) {
+		const repositories = await repositoriesStorage.getAll();
+		/* eslint-disable camelcase */
+		notifications = notifications.filter(({repository: {full_name}}) => {
+			const {owner, repository} = parseFullName(full_name);
+			return repositories[owner][repository];
```
notlmn: This line is throwing an error on me, when I have a notification to be displayed from `fregante/webext-options-sync`.

image

I watch `fregante/webext-options-sync` but do not have access to it, so it would not show up when fetched with `/user/repos`. The same goes for other repos I don't have access to.

source/lib/repositories-service.js
notlmn: CMIIW, shouldn't this be `/user/subscriptions`?

source/repositories.js
```diff
@@ -0,0 +1,156 @@
...
+
+	return `
+		<details>
+			<summary>
+				<label class="${owner}">
+					<input type="checkbox" name="${owner}" />
+					${owner} <span class="count small">(${repos.length})</span>
+				</label>
+			</summary>
+			<div class="list">
```
notlmn: I think there's no need of wrapping this in a `div` here.

notlmn (changes requested):

jroehl:
Is there anything else that needs to be done? :)

notlmn:
I was testing the PR and discovered that we are missing some code all along. Unfortunately, this PR might need some more work. 😅

So the issue is that when we request for notifications, to display them, we have been requesting only the first page. GitHub responds only with the first 50 notifications (even when we use `per_page=1000`), the rest of them are paginated. 

https://github.com/sindresorhus/notifier-for-github/blob/ba11eb909c51d35a7157210ec8acb5a78af1a9bc/source/lib/api.js#L101-L104

For this, the above function requires additional code, parsing the headers too with the awesome function you added in this PR and request for the next pages too. This is required so that if I have any notification filtering set up, and the notifications I need are not in the first page (of JSON response). Then we have to fetch the other pages too, and combine those results with the ones we already fetched until there are no further pages left to fetch.

The PR in its current state works as intended, but only of the notifications to be displayed are not paginated. The issue I'm describing _can_ be considered extreme. WDYT, should we work on that part in this PR or open a new one? I'm gonna leave that up to you @jroehl.

Sorry for extending this PR a little bit more, and thanks for sticking this long with the PR. ❤

notlmn:
Some suggestions meanwhile:
- The `indeterminate` state for checkboxes doesn't work for some reason
- _Can_ disable this entire section when "Show desktop notifications" is unchecked (or mention that this section is somehow related to that field)

jroehl:
So, I managed to fix the issues you mentioned and implemented a recursive function to fetch all the notifications. The issue with the checkboxes was caused by a chrome extension bug with the `chrome_style` option set to `true` in `manifest.json`. 

Could you please check as well if everything works as expected? 🙏

jroehl:
Hi @notlmn - any news on this? :)

source/options.html
```diff
@@ -1,62 +1,90 @@
-<!doctype html>
-<meta charset="utf-8">
+<!DOCTYPE html>
```
sindresorhus: Don't do unrelated changes.

source/options.html
```diff
@@ -1,62 +1,90 @@
...
+<meta charset="utf-8" />
 <title>Notifier for GitHub</title>
-<link rel="stylesheet" href="options.css">
+<link rel="stylesheet" href="options.css" />
```
sindresorhus: Same here and in many other places.

sindresorhus:
> Hi @notlmn - any news on this? :)

@notlmn ^

notlmn:
Sorry folks, this has been a hell of a final year with college and internship going on right now for me.

-------------

Have tested this out locally, working as intended. I suggest @jroehl do a final local test and ping back here so we can merge this.

Error messages are a mess right now, have to fix them later in another PR.

Thanks for you patience everyone.

jroehl:
Hey @notlmn - thx for reverting the unrelated changes. When I tested it again I didn't find any issues. IMHO it's ready to be merged. 🤞

source/lib/api.js
```diff
@@ -93,9 +95,21 @@ export async function getNotificationResponse({maxItems = 100, lastModified = ''
...
+	if (!next) {
+		return notifications;
+	}
+
+	const {searchParams} = new URL(next);
+	return getNotifications({
+		lastModified,
+		page: searchParams.get('page'),
+		maxItems: searchParams.get('per_page')
+	});
```
notlmn: This function call should pass `notifications` as its second argument right? Or am I missing something?

source/lib/api.js
```diff
@@ -93,9 +95,21 @@ export async function getNotificationResponse({maxItems = 100, lastModified = ''
 	return makeApiRequest('/notifications', params);
 }
 
-export async function getNotifications({maxItems, lastModified} = {}) {
-	const {json: notifications} = await getNotificationResponse({maxItems, lastModified});
-	return notifications || [];
+export async function getNotifications({maxItems, lastModified} = {}, notifications = []) {
+	const {headers, json} = await getNotificationResponse({maxItems, lastModified});
```
notlmn: The first argument of this function should accept `page` being sent at https://github.com/sindresorhus/notifier-for-github/pull/192/files#diff-f9e0c5138a3db808bf2f5d5bf33feb49R110, which then again should be passed to `getNotificationResponse`.

This call was never ending because the extension was making requests to the same page (page 1) over and over again until GH API rate limit is exceeded.

The extension made almost 200 requests in just 2 minutes, all to the same page.

image

notlmn (changes requested):
Can you please take a look at my comments. Thanks!

notlmn (approved):

notlmn:
Have tested this as much as I can. Bit I'm sure we will get issues once the change hits other users.

Thanks for the PR @jroehl. Awesome work!

jroehl:
Thanks for taking care of the remaining issues and merging the PR @notlmn 👏

PR #210, opened 2020-04-21
Remove hardcoded protocol for notification page URL

### LINKED ISSUE #198: https should not be hard-coded in the notification page URL
I am using Firefox extension with enterprise github and it is not https-enabled (I know this is bad, but this is what it is). Pretty much everything works, except the opening notifications page when clicking on add-in icon - there is `https` hardcoded in the URL construction: https://github.com/sindresorhus/notifier-for-github/blob/master/source/lib/api.js#L17. Can we somehow use protocol from the `${rootUrl}` provided in add-in settings? Thanks!

Hydrog3n:
Fixes #198 

I created another function who gets the `rootUrl` and processing like removing the last `/` if exist and if `api.github.com` it's will return github website.

Maybe the better way is to create another option?

### DIFF
--- a/source/lib/api.js
+++ b/source/lib/api.js
@@ -8,9 +8,9 @@ export async function getGitHubOrigin() {
 	// TODO: Drop `api.github.com` check when dropping migrations
 	if (origin === 'https://api.github.com' || origin === 'https://github.com') {
 		return 'https://github.com';
-	} else {
-		return origin;
 	}
+
+	return origin;
 }
 
 export async function getTabUrl() {
@@ -27,9 +27,9 @@ export async function getApiUrl() {
 	// TODO: Drop `api.github.com` check when dropping migrations
 	if (origin === 'https://api.github.com' || origin === 'https://github.com') {
 		return origin;
-	} else {
-		return `${origin}/api/v3`;
 	}
+
+	return `${origin}/api/v3`;
 }
 
 export async function getParsedUrl(endpoint, params) {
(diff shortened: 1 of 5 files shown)


source/lib/api.js
```diff
@@ -28,6 +28,16 @@ export async function getApiUrl() {
 	return `${rootUrl}api/v3`;
 }
 
+export async function getRootUrl() {
+	const {rootUrl} = await optionsStorage.getAll();
+
+	if (/(^(https:\/\/)?(api\.)?github\.com)/.test(rootUrl)) {
```
notlmn: I think it's a good time for us to move from regex check here to proper URL parsing

Something like this:

```suggestion
	if ((new URL(rootUrl)).origin === 'https://api.github.com) {
```

notlmn (changes requested):
I like this change, non-destructive for most and we already ask user to include protocol as well in input.

fregante:
You should replace `getHostname` instead of adding a new function; it does the same thing and is used everywhere already.

It should probably be renamed for clarity too

```js
export async function getGitHubOrigin() {
	const {rootUrl} = await optionsStorage.getAll();
	const {origin} = new URL(rootUrl);
	return origin === 'https://api.github.com' ? 'https://github.com' : origin;
}
```

Hydrog3n:
Thank you @notlmn I don't have time to continue my PR sorry.

notlmn:
No problem @Hydrog3n, I am just cleaning up stuff that have been stale for too long. Thanks for you other contributions.

notlmn:
We should actually use `https://github.com` in options for `rootUrl` instead of `https://api.github.com`. Ideally should be done through migrations.

Merging this for now.

PR #214, opened 2020-05-25
Handle dark mode options in Firefox

notlmn:
Long awaited changes from https://github.com/sindresorhus/refined-github/pull/2794.


sindresorhus:
I think we should use https://github.com/fregante/webext-base-css

source/options.css
```diff
@@ -1,4 +1,6 @@
 html {
+	--github-green: #28a745;
+	--github-red: #cb2431;
```
sindresorhus, line 3: Why are you moving this? It's recommended to have variables in `:root`.

source/options.css
```diff
@@ -8,36 +10,31 @@ html.is-edgium {
...
-:root {
-	--github-green: #28a745;
-	--github-red: #cb2431;
-}
-
 body {
 	font-family: system-ui, sans-serif;
 	font-size: 14px;
 	margin: 0;
 	padding: 16px;
```
sindresorhus: I think you can remove many of the properties here. Maybe also in other selectors. Try to remove as much custom CSS as possible.

You can use https://github.com/sindresorhus/refined-github/blob/master/source/options.css as a reference.

sindresorhus:
// @fregante Since it's your lib. In case you see something that could be done better.

source/options.css
notlmn: For some reason Firefox had a grey background in dark mode, so I left these out. Now that's gone, I'll remove this.

notlmn:
Any CSS changes related to repositories filtering may be included in another PR.

Just realized that Firefox doesn't render `<details>` elements in extension options anymore. I could swear it used to work at some point. @fregante, any weird behavior you are seeing in Firefox?

source/options.css
```diff
@@ -1,3 +1,8 @@
+:root {
+	--github-green: #28a745;
+	--github-red: #cb2431;
+}
+
 html {
 	min-width: 550px;
 	overflow-x: hidden;
```
fregante, line 8: `overflow-x` shouldn't be necessary. If it is, it's more likely that some CSS is configured to be wider than it should be (e.g. width: 100% + padding)

fregante: I'd suggest merging these 2 rules into only:

```css
html:not(.is-edgium) {
	min-width: 550px;
}
```

source/options.css
```diff
@@ -31,13 +22,13 @@ h3,
...
 	width: 100%;
 	margin-top: 0;
-	margin-bottom: 0.25em;
+	margin-bottom: 0.25rem;
 }
 
 hr {
 	margin: 1rem -1rem;
 	border: none;
-	border-bottom: solid #d3d3d3 1px;
```
fregante, line 40: https://github.com/fregante/webext-base-css/blob/5e49453e04d8b97eee3fed983877a773b0df7b76/webext-base.css#L43

This whole rule can be dropped. _Maybe_ you only need vertical margins (we don't have them in RGH)

source/options.css
```diff
@@ -31,13 +22,13 @@ h3,
 h4 {
 	width: 100%;
```
fregante, line 14: `width: 100%` is essentially the default for block elements.

source/options.css
```diff
@@ -31,13 +22,13 @@ h3,
...
 	margin-top: 0;
-	margin-bottom: 0.25em;
+	margin-bottom: 0.25rem;
```
fregante, line 16: Unless you're going for very specific spacing, you could drop `margin-bottom` because all of these elements already have it

source/options.css
```diff
@@ -46,38 +37,26 @@ hr {
...
-
-input[type='url'],
-input[type='text'] {
+/*
+Monospace all text inputs
+We don't have many text inputs, so this should be fine
+*/
+input:not([type='checkbox']) {
 	display: block;
 	width: 100%;
```
fregante: Already there: https://github.com/fregante/webext-base-css/blob/5e49453e04d8b97eee3fed983877a773b0df7b76/webext-base.css#L13-L27

source/options.css
```diff
@@ -87,13 +66,15 @@ input:not([type='checkbox']):not(:placeholder-shown):valid {
 
 input:not([type='checkbox']):invalid {
 	color: var(--github-red);
-	border: 1px solid;
+	border: 1px solid var(--github-red);
```
fregante: Can be reverted. `border` uses `color` by default

fregante (changes requested):

webpack.config.js
```diff
@@ -8,20 +9,32 @@ module.exports = {
...
 	plugins: [
+		new MiniCssExtractPlugin(),
 		new SizePlugin(),
 		new CopyWebpackPlugin([
 			{
 				from: '**/*',
 				context: 'source',
-				ignore: ['*.js']
+				ignore: ['*.js', '*.css']
 			},
```
fregante, line 25: If you prefer not having to use `MiniCssExtractPlugin`, you can just set `CopyWebpackPlugin` to copy the raw CSS file like this:
```suggestion
			},
			{
				from: 'node_modules/webext-base-css/webext-base.css'
			}
```

And then just load it in `options.html`
```html
<link rel="stylesheet" href="chrome://global/skin/in-content/common.css">
<link rel="stylesheet" href="webext-base.css">
<link rel="stylesheet" href="options.css">
```

source/options.css
notlmn: Their default top and bottom margin of `1.33em` is too much.

source/options.css
notlmn: We have some `h4` elements inside `label` elements, that have `display: flex` applied. This is to make other content wrap.

source/options.css
notlmn: > `overflow-x` shouldn't be necessary. If it is, it's more likely that some CSS is configured to be wider than it should be (e.g. width: 100% + padding)

That's because of negative margins on `hr` from https://github.com/fregante/webext-base-css/blob/5e49453e04d8b97eee3fed983877a773b0df7b76/webext-base.css#L40-L41.

Body has default margin of `8px`, but `hr` has added with `-17px` margin, more than what is on the `body` (from UA styles).

source/options.css
notlmn: That would be a good addition to `webext-base-css`, WDYT?

source/options.css
fregante: Right, that’s why I suggested dropping that code too. It’s already part of base-css and it doesn’t cause this overflow.

source/options.css
fregante: What? Margins? Horizontal margins are already in base-css. 

Vertical margins no, base-css is just an extension of the default style, not its own framework. Once you start affecting some vertical margins you have to change them all. The default margins work well enough imho

source/options.css
notlmn: @fregante RGH is getting `17px` margin horizontally from somewhere. But this extension is not getting that.

image

I'm getting `8px` as opposed to `17px` in RGH's case. Weird.

image

fregante: Likely because `chrome_style:true` is missing here: https://github.com/sindresorhus/notifier-for-github/blob/dd684a3a69c9e200c2e1df6a6680026b588fc719/source/manifest.json#L36 while RGH has it https://github.com/sindresorhus/refined-github/blob/ef3cee995f8ba8da5337d8220f8ceed80d43febf/source/manifest.json#L31

If that fixes it I'll update base-css’s readme

source/options.css
```diff
@@ -1,28 +1,14 @@
...
+html:not(.is-edgium) {
+	min-width: 550px;
+	overflow-x: hidden;
 }
 
-p {
-	margin-top: 0.25em;
-	margin-bottom: 0;
+body {
 	line-height: 1.5;
```
fregante: If you want to keep this, you can move it to `:root` so this `body` rule can be dropped.

source/options.css
```diff
@@ -46,54 +30,42 @@ hr {
...
-input[type='text'] {
-	display: block;
-	width: 100%;
-	padding: 0.25em;
+/*
+Monospace all text inputs
+We don't have many text inputs, so this should be fine
+*/
+input:not([type='checkbox']) {
+	padding: 0.25rem;
```
fregante: `padding` shouldn't be required. Also the comment seems unnecessary

source/options.css
```diff
@@ -46,54 +30,42 @@ hr {
...
 
 input:not([type='checkbox']):not(:placeholder-shown):valid {
 	color: var(--github-green);
-	border: 1px solid var(--github-green);
+	border: 1px solid;
 }
 
 input:not([type='checkbox']):invalid {
 	color: var(--github-red);
 	border: 1px solid;
```
fregante:
```suggestion
	border-color: var(--github-red) !important;
```

source/options.css
```diff
@@ -46,54 +30,42 @@ hr {
...
+https://github.com/sindresorhus/refined-github/issues/1374#issuecomment-397906701
+*/
 input[type='text'][name='token'] {
-	font-family: Consolas, monospace;
-	font-size: 12px;
-	width: 35ch; /* Show only a part of the token */
+	width: 35ch !important;
 }
 
 .hidden {
```
fregante, line 62: You can use the `hidden` attribute instead

source/options.css
notlmn: > Likely because `chrome_style:true` is missing here:

That indeed seems to be the issue.

source/options.css
notlmn: Was doing that, then realized that it has to be a separate PR (https://github.com/sindresorhus/notifier-for-github/pull/214#issuecomment-634475428). Requires a bit of changes in respective JS files.

source/options.css
notlmn: Reasonable enough, but we do need vertical margins in this case.

source/options.css
notlmn: Dropped that anyway.

fregante (approved):

Issue #220, opened 2020-07-07
Notifier not updating despite having internet connection


connesy:
Version 20.7.6
Despite having internet connection the extension icon still shows "You have to be connected to the internet". When I click on the icon, it works fine and I get taken to github.com/notifications.

![Screenshot from 2020-07-07 09-34-04](https://user-images.githubusercontent.com/13164166/86738305-41534f80-c035-11ea-8fea-8e51668e4580.png)

I have tried closing and opening Google Chrome, disabling and enabling the extension, and turning my internet connection off and on.

notlmn:
This is a long standing issue from Chrome, and is already being tracked by #132. Apparently this is not an issue with Firefox, and seen only in Chrome.

This does seem to be bit different, can you try disabling and re-enabling the extension.

connesy:
That didn't work.

notlmn:
Can you try uninstalling the extension and installing it again, using a fresh personal access token? I'm trying to see why this keeps failing very inconsistently.

connesy:
I've now uninstalled and re-installed the extension using a newly generated personal access token with `repo` and `notifications` permissions. The problem still persists.

notlmn:
Can you help me inspect this by

- Navigating to `chrome://extensions`
- Click on "background views" for this extension
- Check for any errors in console tab

connesy:
I get these errors:

![image](https://user-images.githubusercontent.com/13164166/87013399-ddfb2600-c1ca-11ea-8f00-490b8b05a594.png)

notlmn:
I think that you have "Root URL" option set to `https://github.com` instead of `https://api.github.com`. This limitation would be completely removed when #219 gets resolved. For now you have to use `https://api.github.com`.

connesy:
You are right, that was it. Thanks!

Issue #132, opened 2017-06-08
Notifier stays offline despite regaining connection


benrondeau:
When Chrome goes offline and the extension looses connection to Github, it shows a yellow text bubble saying `off` (or something similar). When you regain internet connection, this bubble stays in place and the only way to reconnect is to manually click on the plugin icon. I have had the plugin be in 'offline' mode for long periods of time and therefore missed notifications.

Ideally, the plugin would re-connect itself and not require the user to click on the plugin icon to reconnect to Github. Not very familiar with the Chrome API but if it does not expose a method to be notified upon regaining internet connection, something like a `setTimeout` that pings the URL every 5 seconds until a connection could be another solution.

@sindresorhus Any thoughts? I will create a PR if you will accept it.

EvertEt:
Should be working since https://github.com/sindresorhus/notifier-for-github-chrome/pull/123 due to the  `window.addEventListener`.

benrondeau:
This is not working for me, unfortunately. I have not investigated specifically why, but it does not work with my setup:

```
Google Chrome Version v58.0.3029.110 (64-bit)
Notifier for GitHub v3.0.3
Mac OSX 10.12.5
```

It will stay offline for as long as I let it. I put my computer to sleep numerous times throughout the day and when I come back, I have to manually reconnect the plugin (clicking on the icon).

benrondeau:
Upgraded to `Google Chrome v59.0.3071.86 (Official Build) (64-bit)` and the problem still exists.

sgrjyb:
Has been updated to https://github.com/sindresorhus/notifier-for-github-chrome/archive/3.0.4.zip

benrondeau:
@sindresorhus Some issues with 3.0.4 release:

- [ ] You didn't update the manifest version from `3.0.3` to `3.0.4`
- [ ] This extension seems to have issues is Chrome 56 (see image below)

<img width="584" alt="screen shot 2017-06-08 at 11 35 02 pm" src="https://user-images.githubusercontent.com/5001219/26961300-5e355eda-4ca3-11e7-9a85-d31cbfe4beac.png">

- [ ] Manually installing 3.0.4 gives this error:

<img width="726" alt="screen shot 2017-06-08 at 11 35 25 pm" src="https://user-images.githubusercontent.com/5001219/26961308-70682f7e-4ca3-11e7-90db-61b383f2bed3.png">

gcochard:
[...]

PR #221, opened 2020-07-07
Use persistent background pages

notlmn:
On Chrome we had this long standing issue where the extension goes offline until you click on the extension icon (#132) or trigger background pages some other way. In some cases it goes completely unresponsive despite clicking on the extension icon (#220).

We haven't had similar issues on Firefox, which does not (yet) support non-persistent background pages, i.e., the background page stays loaded all the time.

https://developer.chrome.com/extensions/background_pages

Chrome suggests not to have resource intensive tasks in background pages, but all we have in background pages is an alarm, and a couple of network requests, so this should not be a problem.

I'm hoping changing to the same persistent model as Firefox on Chrome should fix issues mentioned above.

/cc @fregante and @sindresorhus for opinions.

### DIFF
--- a/source/background.js
+++ b/source/background.js
@@ -19,7 +19,8 @@ async function scheduleNextAlarm(interval) {
 	// Delay less than 1 minute will cause a warning
 	const delayInMinutes = Math.max(Math.ceil(intervalValue / 60), 1);
 
-	browser.alarms.create({delayInMinutes});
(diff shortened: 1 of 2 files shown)


sindresorhus:
Yeah, let's try it out.

fregante:
Did it fix it? The issue is still open

PR #249, opened 2021-04-06
Remove redundant regex for token check

### LINKED ISSUE #248: Error: missing token (new token format)
After updating to 21.4.6 (Firefox) we can set the new token but it cannot be used due to this check:

https://github.com/sindresorhus/notifier-for-github/blob/c9c66fe81df2767eb8ec705cba3d3f066728583b/source/lib/api.js#L44-L46

RafaelKr:
fix #248

https://regex101.com/r/yMMKp9/1/

### DIFF
--- a/source/lib/api.js
+++ b/source/lib/api.js
@@ -41,10 +41,6 @@ export async function getParsedUrl(endpoint, params) {
 export async function getHeaders() {
 	const {token} = await optionsStorage.getAll();
 
-	if (!(/ghp_\w{36,251}|[a-z\d]{40}/.test(token))) {
-		throw new Error('missing token');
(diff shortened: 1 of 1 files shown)


fregante:
I think this check is redundant and should be removed. Invalid tokens are not saved

RafaelKr:
cc @sindresorhus I can remove this check at all if you want?

Edit: I think it makes sense. Already removed it.

fregante:
You might want to  keep it as a plain existence check. `if !token, throw`

RafaelKr:
Updated. Thanks!

fregante (approved):

PR #252, opened 2021-04-06
Replace webpack with parcel

fregante:
I was trying to fix []() https://github.com/sindresorhus/notifier-for-github/issues/251 but of course webpack’s config was a PITA to update and conform to ESM, so the next logical step was… https://github.com/fregante/browser-extension-template/pull/44

To get started I ran:

```
npm i -D parcel@nightly @parcel/config-webextension@nightly
```

Issue #251, opened 2021-04-06
Update tests to use AVA 3


fregante:
Config can be copied from https://github.com/sindresorhus/refined-github/blob/main/package.json

The `esm` dependency should be dropped and I think the package type should be changed to `module`, either via env or in `package.json` (the latter would probably require extra changes though)

More details might be found in https://github.com/avajs/ava/issues/2293

fregante:
WIP

PR #253, opened 2021-04-07
Lint and update dependencies

fregante:
Will help fix []() #251 

Follows and is based on #252

Untested but it looks equivalent

### DIFF
--- a/source/lib/repositories-service.js
+++ b/source/lib/repositories-service.js
@@ -36,7 +36,7 @@ export async function listRepositories(update) {
 		for (const {full_name} of fetched) {
 			const {owner, repository} = parseFullName(full_name);
 			tree[owner] = tree[owner] || {};
-			tree[owner][repository] = Boolean(stored?.[owner]?.[repository]);
+			tree[owner][repository] = Boolean(stored && stored[owner] && stored[owner][repository]);
 		}
 
 		/* eslint-enable camelcase */
--- a/source/lib/repositories-service.js
+++ b/source/lib/repositories-service.js
@@ -1,15 +1,17 @@
-import {parseLinkHeader, parseFullName} from '../util';
-import repositoriesStorage from '../repositories-storage';
-import {makeApiRequest} from './api';
+import {parseLinkHeader, parseFullName} from '../util.js';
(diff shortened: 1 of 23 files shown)


source/lib/repositories-service.js
```diff
@@ -27,18 +29,17 @@ export async function getRepositories(
...
-		tree = fetched.reduce((tree, {full_name}) => {
+		for (const {full_name} of fetched) {
 			const {owner, repository} = parseFullName(full_name);
-			return Object.assign({}, tree, {
-				[owner]: Object.assign(tree[owner] || {}, {
-					[repository]: Boolean(stored && stored[owner] && stored[owner][repository])
-				})
+			tree[owner] = Object.assign(tree[owner] || {}, {
+				[repository]: Boolean(stored && stored[owner] && stored[owner][repository])
 			});
```
fregante: .reduce brought us to do unspeakable things

```suggestion
			tree[owner] = tree[owner] || {};
			tree[owner][repository] = Boolean(stored && stored[owner] && stored[owner][repository]);
```

Or this?

```suggestion
			tree[owner] = tree[owner] || {};
			tree[owner][repository] = Boolean(stored?.[owner]?.[repository]);
```

Reading the code, however, it seems that this code is wrong in many parts. `stored` is never null nor empty since it gets the default options.

sindresorhus: The latter

test/util-test.js
```diff
@@ -32,7 +32,7 @@ test.serial('isChrome validates User-Agent string', t => {
 
 test.serial('isNotificationTargetPage returns true for only valid pages', async t => {
 	// Invalid pages
-	t.throwsAsync(() => isNotificationTargetPage(''));
+	await t.throwsAsync(() => isNotificationTargetPage(''));
```
sindresorhus:
```suggestion
	await t.throwsAsync(isNotificationTargetPage(''));
```

sindresorhus:
Should we make the package `type: module` while we're at it?

source/lib/repositories-service.js
fregante: It'll have to be done separately because this branch is on Node 11

PR #254, opened 2021-04-07
Update AVA

### LINKED ISSUE #251: Update tests to use AVA 3

fregante:
Fixes #251 

Follows and is based on #253

### DIFF
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -190,7 +190,9 @@ test.serial('#makeApiRequest returns rejected promise for 4xx status codes', asy
 		statusText: 'Not found'
 	});
 
-	await t.throwsAsync(() => service.makeApiRequest('notifications'), 'client error');
+	await t.throwsAsync(service.makeApiRequest('notifications'), {
+		message: 'client error'
+	});
 });
 
 test.serial('#makeApiRequest returns rejected promise for 5xx status codes', async t => {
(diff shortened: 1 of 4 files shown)


test/api-test.js
```diff
@@ -190,7 +190,9 @@ test.serial('#makeApiRequest returns rejected promise for 4xx status codes', asy
 		statusText: 'Not found'
 	});
 
-	await t.throwsAsync(() => service.makeApiRequest('notifications'), 'client error');
+	await t.throwsAsync(() => service.makeApiRequest('notifications'), {
```
sindresorhus: Remove the `() => ` part.

fregante: How can `t.throwsAsync` catch and compare the error if the call throws before its result is passed to `t.throwsAsync`? 🤔

sindresorhus: It doesn't throw. `makeApiRequest` returns a promise. You pass a promise to `throwsAsync` and it checks whether it rejects or not.

fregante: Promises, how do they work

fregante:
@notlmn It turns out most of the changes required for ESM support were done by XO in #253 

It pays to update the dependencies!

notlmn:
@fregante thank you for this while I am MIA out of OSS!

PR #267, opened 2022-03-01
Only register the notification click handler once

### LINKED ISSUE #234: Open 10 tabs when clicking notification in Mac OS Big Sur / Firefox
MacOS open 10 new tab for the same page when I receive and clicking on a notification,

OS : Mac OS Big Sur
Browser : Firefox.

any idea ?

mrmr1993:
This PR replaces the handler for `browser.notification.onClicked.addListener` with a named function, which avoids re-registering the handler on every call to `addHandlers`.

Note that `addEventListener` called repeatedly with the same arguments will not register the same event listener multiple times (see the [`EventTarget.addEventListener` spec](https://www.w3.org/TR/2000/REC-DOM-Level-2-Events-20001113/events.html#Events-EventTarget-addEventListener) for more info). This means that using the same named function for each call to `addHandlers` will only call that function once, where the previous code added a new anonymous function for each call.

This fixes #234, where clicking a notification would result in multiple tabs being opened, depending on the number of times that `addHandlers` was called.

### DIFF
--- a/source/background.js
+++ b/source/background.js
@@ -97,13 +97,15 @@ async function onTabUpdated(tabId, changeInfo, tab) {
 	}
 }
 
+function openNotificationSync(id) {
+    openNotification(id);
+}
+
 async function addHandlers() {
 	const {updateCountOnNavigation} = await optionsStorage.getAll();
 
 	if (await queryPermission('notifications')) {
-		browser.notifications.onClicked.addListener(id => {
-			openNotification(id);
-		});
+		browser.notifications.onClicked.addListener(openNotificationSync);
 	}
 
 	if (await queryPermission('tabs')) {
--- a/source/background.js
+++ b/source/background.js
@@ -98,7 +98,7 @@ async function onTabUpdated(tabId, changeInfo, tab) {
 }
 
 function openNotificationSync(id) {
-    openNotification(id);
+	openNotification(id);
 }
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Linting is failing

source/background.js
```diff
@@ -97,13 +97,15 @@ async function onTabUpdated(tabId, changeInfo, tab) {
 	}
 }
 
+function openNotificationSync(id) {
```
notlmn:
```suggestion
function onNotificationClick(id) {
```

source/background.js
```diff
@@ -97,13 +97,15 @@ async function onTabUpdated(tabId, changeInfo, tab) {
...
+}
+
 async function addHandlers() {
 	const {updateCountOnNavigation} = await optionsStorage.getAll();
 
 	if (await queryPermission('notifications')) {
-		browser.notifications.onClicked.addListener(id => {
-			openNotification(id);
-		});
+		browser.notifications.onClicked.addListener(openNotificationSync);
```
notlmn:
```suggestion
		browser.notifications.onClicked.addListener(onNotificationClick);
```

notlmn (changes requested):
Awesome, thank you for the fix (except for some nits).

notlmn (approved):
LGTM. Thanks again!

PR #277, opened 2023-03-04
Fix calling toGMTString on lastModified

### LINKED ISSUE #276: Desktop notifications are not sent
Extension version: 23.2.15
Firefox version: 110.0
OS: Manjaro Linux. Kernel: 6.0.19

This is happening recently. Everything is working fine and I can see notifications count in the icon. Only problem is with sending desktop notifications, even though it is enabled in the settings.

I thought this is an issue with Firefox or my current configuration, but then I tested sending a desktop notification using this extension (https://addons.mozilla.org/en-US/firefox/addon/notification-test/) and it is working fine.

mabrikan:
This fixes #276.

Function `checkNotifications` is always passed a *string* value for `lastModified` (unless, `localStore` is empty, in which case a `Date` *object* is created), so calling `toGMTString()` resulted in an error.

This change does the following:

1. Fix the bug by removing unneeded `toGMTString()` call
2. make sure `lastModified` is always a *string*, even when `localStore` is empty
3. Use `toUTCString()` as it is more consistent with how JS interprets time.

### DIFF
--- a/source/background.js
+++ b/source/background.js
@@ -25,9 +25,9 @@ async function scheduleNextAlarm(interval) {
 }
 
 async function handleLastModified(newLastModified) {
-	const lastModified = await localStore.get('lastModified') || new Date(0);
+	const lastModified = await localStore.get('lastModified') || new Date(0).toUTCString();
 
-	// Something has changed since we last accessed, display any new notificaitons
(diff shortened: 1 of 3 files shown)


notlmn:
Thank you 🙌

mabrikan:
You're most welcome.

PR #279, opened 2023-03-09
Add sanity checks for permissions

### LINKED ISSUE #278: Random notification chime every minute or so
Starting this morning I've had random notification chimes play, without any notification showing, around every minute or so. I have tried revoking notification permissions as well as muting every open tab until I found the culprit, being Notifier for Github. Unchecking the "Enable notification sound" checkbox in the extension settings allows me to not have the sound, but I'd like to have it when there's a legitimate notification.

I'm using Chrome Version 111.0.5563.65 (Build officiel) (64 bits) on Windows 10.

I have seen [this issue](https://github.com/sindresorhus/notifier-for-github/issues/255) and [that fix](https://github.com/sindresorhus/notifier-for-github/pull/275) but I guess it broke again ?

Since it seems related to the locale & time, I am located in Paris, France, using a french locale.

mabrikan:
Fixes #278.

`browser.notifications.create` is `undefined` when notification permission is _not granted_.

This makes calls to function `showNotifications` fail stopping the program flow and preventing saving `lastModified` field.

Also perhaps we need to unset permission-related options whenever a permission is removed.
`onRemove`[1] listener is a good approach.

[1]
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/onRemoved

### DIFF
--- a/source/lib/notifications-service.js
+++ b/source/lib/notifications-service.js
@@ -7,6 +7,7 @@ import {makeApiRequest, getNotifications, getTabUrl, getGitHubOrigin} from './ap
 import {getNotificationReasonText} from './defaults.js';
 import {openTab} from './tabs-service.js';
 import localStore from './local-store.js';
+import {queryPermission} from './permissions-service.js';
 
 function getLastReadForNotification(notification) {
 	// Extract the specific fragment URL for a notification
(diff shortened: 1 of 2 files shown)


notlmn:
> Also perhaps we need to unset permission-related options whenever a permission is removed. `onRemove`[1] listener is a good approach.
> 
> [1] [developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/onRemoved](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/onRemoved)

I'm not sure if we can keep the options storage and permissions from the browser side in sync. The best approach as you have documented here is to always check for permissions before performing any action.

notlmn (approved):

notlmn:
Thanks for the CL 🙌

PR #281, opened 2023-05-25
Updating auth header to use Bearer

housekelly:
Not able to get this to work on latest GHES instance and noticed it's using 'token' as the auth header.  Tested in postman and works only using 'Bearer'.

https://docs.github.com/en/enterprise-server@3.8/rest/overview/authenticating-to-the-rest-api#about-authentication


housekelly:
@sindresorhus any issues with the above ?

PR #298, opened 2024-07-20
Migrate to Extension Manifest v3

### LINKED ISSUE #297: URGENT: Manifest v3
# **Urgent**

The extension will soon be unusable on Chromium due to the deprecation of Manifest v2

image

It needs to make the transition to Manifest v3

notlmn:
Fixes: #297

- Move background to a service worker
- Move audio playback to offscreen file
- Convert script types to modules in options and offscreen
- Update minimum browser versions for Chrome and Firefox
- Update dev dependencies as needed
- Update tests as needed (remove some individual tests)
- Disable `badge-test` as `sinon-chrome` does not have stub for `browser.action` API
- Disable auto-publishing to AMO due to lack of offscreen API in Firefox

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         command:
-          - firefox
+          # - firefox
           - chrome
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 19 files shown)


Malix-off:
Closes #297

notlmn:
@sindresorhus can you do a manual release to CWS from this branch instead of the default one? Or do you have any other suggestions for working around v2 and v3 releases.

This one fails to run for Firefox because there's no offscreen API in Firefox yet.

sindresorhus:
I think it would be better to just disable Firefox auto-publishing for now: https://github.com/sindresorhus/notifier-for-github/blob/main/.github/workflows/release.yml

PR #303, opened 2025-01-20
Remove "api." suffix from default rootUrl

Flamefire:
Match the description for (e.g.) new installations, see https://github.com/sindresorhus/notifier-for-github/blob/7aade100e6b4d3745406deee7bfe5d36c30353ea/source/options.html#L15


notlmn:
Thanks!

PR #316, opened 2026-06-12
Fix offline recovery after reconnect

landeqiming666:
## Summary
- Restore recovery from offline state when the browser reports it is online again.
- Keep the connection helper from staying stuck in the offline state after a transient network loss.
- Add regression coverage for both the connection helper and background recovery path.

## Validation
- `PATH=/Users/lijiaxuan/MyInformation/.local-node22/bin:$PATH npm run test:js -- test/connection-test.js test/background-test.js`
- `PATH=/Users/lijiaxuan/MyInformation/.local-node22/bin:$PATH npm run lint:js`
- `PATH=/Users/lijiaxuan/MyInformation/.local-node22/bin:$PATH npm run lint:css`
- `PATH=/Users/lijiaxuan/MyInformation/.local-node22/bin:$PATH npm run build`
- `PATH=/Users/lijiaxuan/MyInformation/.local-node22/bin:$PATH npm test`
- `git diff --check`

Fixes #132.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#132: Notifier stays offline despite regaining connection](https://oss.issuehunt.io/repos/18894771/issues/132)
---
</details>

PR #317, opened 2026-06-27
[codex] Fix offline recovery after reconnect

### LINKED ISSUE #132: Notifier stays offline despite regaining connection

bpc-oss:
## Summary
- restore offline recovery in the Manifest V3 background worker by listening for browser connection changes
- treat API `network error` failures as offline-state transitions so the badge recovers cleanly after reconnect
- add regression coverage for the reconnect path and harden the reconnect assertions and network-error matching

## Root Cause
The current MV3 background worker can render the offline badge, but it does not actively recover when connectivity returns. It also routes fetch-level network failures through a generic error path instead of re-entering offline recovery.

## Validation
- `npx -y -p node@22 -p npm@10 npm test`
- `git diff --check`

Fixes #132.


Slicksz1306:
Thanks, this looks like the best candidate for fixing #132.

I compared it with the other open reconnect PRs. This version keeps the scope tight, preserves the existing alarm fallback, adds regression coverage for the background recovery path, and handles network-error matching a little more defensively than #316.

CI is green (`npm test`). I don’t see a blocking issue in the patch. The only contextual caveat is that online/offline events in worker contexts have some browser-compatibility nuance, but since the existing alarm polling remains in place, this looks acceptable as an incremental fix.

Once this is marked ready for review, I think it can supersede #313, #314, #315, and #316.

